PR #13, opened 2024-07-03
feat: Add a vec of trait datasources to historical data fetcher

ferranbt:
## 📝 Summary

This PR removes the hardcoded data sources in the HistoricalDataFetcher and replaces them with a dynamic Datasource trait.

## 💡 Motivation and Context

Before, the two main data sources (mempool dumpster, fb relay) were hardcoded in the HistoricalDataFetcher service. Now, the structure is more flexible and new data sources can be plugged into the HistoricalDataFetcher without any changes to it.

---

* [x] Run `make lint`
* [x] Run `make test`

### DIFF
--- a/crates/rbuilder/src/backtest/fetch/datasource.rs
+++ b/crates/rbuilder/src/backtest/fetch/datasource.rs
@@ -1,16 +1,16 @@
 use crate::backtest::OrdersWithTimestamp;
 use async_trait::async_trait;
 
-/// Datasource trait
+/// DataSource trait
 ///
 /// This trait is used to fetch data from a datasource
 #[async_trait]
-pub trait Datasource: std::fmt::Debug {
-    async fn get_data(&self, block: BlockRef) -> eyre::Result<Vec<OrdersWithTimestamp>>;
-    fn clone_box(&self) -> Box<dyn Datasource>;
+pub trait DataSource: std::fmt::Debug {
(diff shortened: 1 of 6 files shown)


metachris (approved):

crates/rbuilder/src/backtest/fetch/datasource.rs
```diff
@@ -0,0 +1,32 @@
+use crate::backtest::OrdersWithTimestamp;
+use async_trait::async_trait;
+
+/// Datasource trait
+///
+/// This trait is used to fetch data from a datasource
+#[async_trait]
+pub trait Datasource: std::fmt::Debug {
+    async fn get_data(&self, block: BlockRef) -> eyre::Result<Vec<OrdersWithTimestamp>>;
+    fn clone_box(&self) -> Box<dyn Datasource>;
```
dvush: Why this instead of making Datasource Clonable (the same way it has std::fmt::Debug)?

ferranbt: I tried it but it does not work. Do not have enough context in Rust to know. This was generated by Claude but also a common pattern I found for this use case of storing vec<box<dyn Trait>>>

ZanCorDX: dyn can't be used with Clonable since it's Sized :(.

ZanCorDX (changes requested):
We use Source with S upper case in several places can we call it DataSource?
get_data -> get_orders?

ZanCorDX (approved):

PR #33, opened 2024-07-06
fix c-kzg dependency version

fiiiu:
## 📝 Summary

reth dependencies use `c-kzg` version `1.0.2`, while `ethereum-consensus` requires version `1.0.0`. This fixes the version of `c-kzg` to `1.0.0` which works for all dependencies. Note that this is a temporary fix since when we move to reth version `1.0.0` it will require `c-kzg` version `1.0.2`.


fiiiu:
Closing, this was "magically" fixed somewhere between `7803c8f` and `b2cb717`.

PR #34, opened 2024-07-06
feat: Add abstraction for the beacon api client

ferranbt:
## 📝 Summary

This PR adds a small abstraction over the beacon client api and implements the methods required by the rbuilder.

There are a couple of unit tests that require a running beacon node. They are ignore right now but we will enable them in a followup PR.

## 💡 Motivation and Context

---

* [x] Run `make lint`
* [x] Run `make test`

### DIFF
--- a/crates/rbuilder/Cargo.toml
+++ b/crates/rbuilder/Cargo.toml
@@ -71,6 +71,7 @@ secp256k1 = { version = "0.27.0", features = [
 ] }
 rayon = "1.8.0"
 flate2 = "1.0.27"
+# Version required by ethereum-consensus beacon-api-client
 mev-share-sse = { git = "https://github.com/paradigmxyz/mev-share-rs", rev = "9eb2b0138ab3202b9eb3af4b19c7b3bf40b0faa8", default-features = false }
 jsonrpsee = { version = "0.20.3", features = ["full"] }
 ssz_rs = { git = "https://github.com/ralexstokes/ssz-rs.git", version = "0.9.0" }
--- a/crates/rbuilder/Cargo.toml
+++ b/crates/rbuilder/Cargo.toml
@@ -71,9 +71,10 @@ secp256k1 = { version = "0.27.0", features = [
 ] }
 rayon = "1.8.0"
 flate2 = "1.0.27"
-mev-share-sse = "0.1.5"
+mev-share-sse = { git = "https://github.com/paradigmxyz/mev-share-rs", rev = "9eb2b0138ab3202b9eb3af4b19c7b3bf40b0faa8", default-features = false }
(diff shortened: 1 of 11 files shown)


crates/rbuilder/Cargo.toml
```diff
@@ -71,9 +71,10 @@ secp256k1 = { version = "0.27.0", features = [
 ] }
 rayon = "1.8.0"
 flate2 = "1.0.27"
-mev-share-sse = "0.1.5"
+mev-share-sse = { git = "https://github.com/paradigmxyz/mev-share-rs", rev = "9eb2b0138ab3202b9eb3af4b19c7b3bf40b0faa8", default-features = false }
```
ryanschneider, line 75: Suggestion: add a PR comment explaining why any crate dependency updates were necessary.

ryanschneider (approved):
lgtm! Though would be good to document somewhere why we had to point to a random `mev-share-sse` commit.

crates/rbuilder/Cargo.toml
dvush: agree, so we know when to remove it

ferranbt: Good point, so when I tried to use `https://github.com/ralexstokes/ethereum-consensus` for the api It needed to lock this specific version.

crates/rbuilder/src/beacon_api_client/mod.rs
```diff
@@ -0,0 +1,88 @@
...
+impl Debug for Client {
+    fn fmt(&self, f: &mut std::fmt::Formatter<'_>) -> std::fmt::Result {
+        f.debug_struct("Client").finish()
+    }
+}
+
+impl Default for Client {
+    fn default() -> Self {
+        Self {
+            inner: bClient::new(Url::parse("http://localhost:8000").unwrap()),
```
ZanCorDX: don't use hidden unwraps

ZanCorDX: can you put "http://localhost:8000" as a const and replace it in the 3 places you use it?

crates/rbuilder/Cargo.toml
ZanCorDX: Can we add the comment on the toml just above this line?

crates/rbuilder/src/live_builder/base_config.rs
```diff
@@ -242,6 +244,16 @@ impl BaseConfig {
         genesis_value_parser(&self.chain)
     }
 
+    pub fn beacon_clients(&self) -> Vec<Client> {
```
ZanCorDX: remove hidden unwrap

PR #35, opened 2024-07-06
feat: Derive builder domain from beacon api

ferranbt:
## 📝 Summary

Builds on top of #34.

This PR removes the flag `cl_network_config_dir` which is used to load the consensus network spec to derive the builder signing domain. Instead, it uses the beacon clients to retrieve the required fields.

---

* [x] Run `make lint`
* [x] Run `make test`


crates/rbuilder/src/live_builder/base_config.rs
```diff
@@ -590,29 +608,42 @@ pub fn coinbase_signer_from_secret_key(secret_key: &str) -> eyre::Result<Signer>
...
-        })?;
+fn get_signing_domain(chain: Chain, client: Client) -> eyre::Result<B256> {
+    let cl_context = match chain.kind() {
+        ChainKind::Named(NamedChain::Mainnet) => ContextEth::for_mainnet(),
+        ChainKind::Named(NamedChain::Sepolia) => ContextEth::for_sepolia(),
+        ChainKind::Named(NamedChain::Goerli) => ContextEth::for_goerli(),
+        ChainKind::Named(NamedChain::Holesky) => ContextEth::for_holesky(),
+        _ => {
+            let rt = Runtime::new()?;
+            let spec = rt.block_on(client.get_spec())?;
```
ryanschneider: Maybe Handle::block_on is more idiomatic and can be wrapped in task::block_in_place? https://docs.rs/tokio/latest/tokio/task/fn.block_in_place.html#examples

ferranbt: Cannot make it work. I think there is some nuance there that makes these two functions different.

crates/rbuilder/src/live_builder/base_config.rs
```diff
@@ -602,29 +602,46 @@ pub fn coinbase_signer_from_secret_key(secret_key: &str) -> eyre::Result<Signer>
...
-            eyre::eyre!("Unknown chain. If network is not named, specify cl_network_config_dir")
-        })?;
+fn get_signing_domain(chain: Chain, beacon_clients: Vec<Client>) -> eyre::Result<B256> {
+    let cl_context = match chain.kind() {
+        ChainKind::Named(NamedChain::Mainnet) => ContextEth::for_mainnet(),
+        ChainKind::Named(NamedChain::Sepolia) => ContextEth::for_sepolia(),
+        ChainKind::Named(NamedChain::Goerli) => ContextEth::for_goerli(),
+        ChainKind::Named(NamedChain::Holesky) => ContextEth::for_holesky(),
+        _ => {
+            let rt = Runtime::new()?;
```
ZanCorDX: It's illegal to create nested runtimes, were you able to run this from a tokio main?

ZanCorDX (changes requested):

crates/rbuilder/src/live_builder/base_config.rs
ferranbt: I was able to run this from a tokio::test. Do you think I can use something like [this](https://github.com/flashbots/rbuilder/pull/35#discussion_r1667393920).

ZanCorDX: The final solution is to admit that this is a blocking call and propagate the async, it's not nice to block too much on tokio unless you are on a tokio::spawn_blocking or Godzilla will come out and kick your ass.

ZanCorDX (approved):

PR #36, opened 2024-07-06
Add proc macros to ignore tests if env or http not set

ferranbt:
## 📝 Summary

This PR creates two new procedural macro (and crate otherwise it does not work) that:
- Ignore a test if some env variable is not set.
- Ignore a test if some endpoint is not available.

---

* [x] Run `make lint`
* [x] Run `make test`


ryanschneider (approved):
Great idea, lgtm!

dvush (approved):

PR #37, opened 2024-07-06
Add a trait to abstract the access to the state

ferranbt:
## 💡 Motivation and Context

---

* [ ] Run `make lint`
* [ ] Run `make test`


crates/rbuilder/src/live_builder/config.rs
```diff
@@ -378,80 +372,29 @@ impl Default for Config {
     }
 }
 
-/// Open reth db and DB should be opened once per process but it can be cloned and moved to different threads.
-pub fn create_provider_factory(
```
ferranbt, line 382: This code was repeated in `utils.rs`.

PR #38, opened 2024-07-07
Fix: Use lock properly in filter_order_by_nonces to do it concurrently

ghost:
## 📝 Summary

The current use of Mutex on `nonce_cache` in `filter_order_by_nonces` is not correct. It causes the function to run sequentially because every other task will wait for prev one ( both read and write ) complete in order to do its work.
I change to use RwLock instead. Before the change it takes 18m to fetch a block with 2k4 orders, and now it takes 18s with 400 concurrent request. 

- And lint the file

## 💡 Motivation and Context

---

* [x] Run `make lint`
* [x] Run `make test`

### DIFF
--- a/crates/rbuilder/src/backtest/fetch/mod.rs
+++ b/crates/rbuilder/src/backtest/fetch/mod.rs
@@ -1,8 +1,16 @@
+pub mod datasource;
 pub mod flashbots_db;
 pub mod mempool;
 pub mod mev_boost;
 
-use crate::{backtest::BlockData, mev_boost::BuilderBlockReceived, utils::timestamp_as_u64};
+use crate::{
+    backtest::{
+        fetch::datasource::{BlockRef, DataSource},
+        BlockData,
+    },
+    mev_boost::BuilderBlockReceived,
+    utils::timestamp_as_u64,
+};
 
 use alloy_provider::Provider;
 use alloy_rpc_types::{Block, BlockId, BlockNumberOrTag};
@@ -29,9 +37,7 @@ use crate::{
 pub struct HistoricalDataFetcher {
     eth_provider: BoxedProvider,
     eth_rpc_parallel: usize,
-    mempool_datadir: PathBuf,
-    // If none, skip bundles
-    flashbots_db: Option<PgPool>,
+    data_sources: Vec<Box<dyn DataSource>>,
     payload_delivered_fetcher: PayloadDeliveredFetcher,
 }
 
@@ -42,13 +48,24 @@ impl HistoricalDataFetcher {
         mempool_datadir: PathBuf,
         flashbots_db: Option<PgPool>,
     ) -> Self {
+        let mut data_sources: Vec<Box<dyn DataSource>> = vec![Box::new(
+            mempool::MempoolDumpsterDatasource::new(mempool_datadir),
+        )];
+
+        if let Some(db_pool) = flashbots_db {
+            data_sources.push(Box::new(RelayDB::new(db_pool)));
+        }
+
         Self {
             eth_provider,
             eth_rpc_parallel,
-            mempool_datadir,
-            flashbots_db,
+            data_sources,
             payload_delivered_fetcher: PayloadDeliveredFetcher::default(),
         }
+    }
+
+    pub fn add_datasource(&mut self, datasource: Box<dyn DataSource>) {
+        self.data_sources.push(datasource);
     }
 
     async fn get_payload_delivered_bid_trace(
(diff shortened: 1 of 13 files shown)


crates/rbuilder/src/backtest/fetch/mod.rs
```diff
@@ -176,29 +180,33 @@ impl HistoricalDataFetcher {
                 async move {
                     let mut all_nonces_failed = true;
                     for nonce in nonces {
-                        let onchain_nonce = {
-                            let mut nonce_cache = nonce_cache.lock().await;
+                        let mut res_onchain_nonce = 0_u64;
```
bertmiller: Would it make more sense to us an `Option<u64>` instead? Imagine that the nonce_cache is initialized and has a value of `0` for this address already. Then we hit the `if res_onchain_nonce == 0` codepath and repeat the nonce call path even tho we already know the appropriate nonce value. It's not the worst thing in the world, but there is perhaps some more optimization you could do!

bertmiller (commented):
Thanks for this PR! Really appreciate it and good catch :)

I think there is a potential race condition in the nonce cache update logic. Currently, the code checks if a value exists, fetches it if not, then writes it back. This could lead to overwriting more recent values in concurrent scenarios: imagine if there were two transactions from account X in different threads.

Here's a scenario that illustrates the problem:
- Thread A checks the cache for address X, doesn't find it
- Thread B checks the cache for address X, doesn't find it
- Thread A fetches nonce N1 from the network for address X
- Thread B fetches nonce N2 from the network for address X (N2 > N1 because it's more recent)
- Thread B writes N2 to the cache
- Thread A writes N1 to the cache, overwriting the more recent N2

What do you think about using a single write lock operation that checks and updates atomically? Something like this:

```rust
let mut nonce_cache = nonce_cache.write().map_err(|e| eyre::eyre!("Failed to acquire write lock: {}", e))?;
if !nonce_cache.contains_key(&nonce.address) {
    let address = nonce.address;
    let onchain_nonce = self
        .eth_provider
        .get_transaction_count(address)
        .block_id(BlockId::Number(parent_block.into()))
        .await
        .wrap_err("Failed to fetch onchain tx count")?;
    nonce_cache.insert(nonce.address, onchain_nonce);
}
let res_onchain_nonce = *nonce_cache.get(&nonce.address).unwrap();
```

ghost:
@bertmiller spot on the analysis that there is an overwrite on the nonce value for the same address X. However, the part of getting the nonce, is a nonce of an address from parent block which should be the same regardless of the order of onchain network request to get the data.

dvush (approved):

ZanCorDX:
Thanks!!!!! 
I believe the change from Mutex to RWlock is unimportant, but the **great catch** is not keeping the lock while calling get_transaction_count.
Can you remove the is_none and the 2 unwraps in favor of an if let?

ghost:
> Thanks!!!!! I believe the change from Mutex to RWlock is unimportant, but the **great catch** is not keeping the lock while calling get_transaction_count. Can you remove the is_none and the 2 unwraps in favor of an if let?

Thanks for your feedback. At the `is_none()` checking, we don't intend to read the value, only write if it's none and for the 2 `unwrap()`, the value surely is available, so I don't think we need to use if-let here, I believe.

ZanCorDX:
> > Thanks!!!!! I believe the change from Mutex to RWlock is unimportant, but the **great catch** is not keeping the lock while calling get_transaction_count. Can you remove the is_none and the 2 unwraps in favor of an if let?
> 
> Thanks for your feedback. At the `is_none()` checking, we don't intend to read the value, only write if it's none and for the 2 `unwrap()`, the value surely is available, so I don't think we need to use if-let here, I believe.

It's just to make the code more robust, in most cases rust allows us to avoid wrap() which might fail if something changes in the future.
It would be like this:
```
                     let res_onchain_nonce = if let Some(res_onchain_nonce) = res_onchain_nonce {
                            res_onchain_nonce
                        } else {
                            let address = nonce.address;
                            let onchain_nonce = self
                                .eth_provider
                                .get_transaction_count(address)
                                .block_id(BlockId::Number(parent_block.into()))
                                .await
                                .wrap_err("Failed to fetch onchain tx count")?;

                            if let Ok(mut nonce_cache) = nonce_cache.write() {
                                nonce_cache.entry(address).or_insert(onchain_nonce);
                            }
                            onchain_nonce
                        };
                        if res_onchain_nonce > nonce.nonce && !nonce.optional {
```
This way the code enforces the correctness at compile time and not at runtime just by the logic of it.

ghost:
it makes sense.

PR #59, opened 2024-07-13
feat: Add the ability to read the cl_node_urls from env variables

MoeMahhouk:
## 📝 Summary

This PR adds the ability to parse the cl_node_url passed from an environment variable too.
It would help with the automatic deployment in the TDX yocto images.

This addition does not affect the original behaviour of accepting a single value or an array of values. It adds extra a way to pass the values from an env variable where the values are separated by a comma from a single string.

## 💡 Motivation and Context

The motivation behind this change was triggered when preparing the TDX demo with rbuilder and the rbuilder was crashing because we were passing the CL_NODE_URL as environment variables during deployment from the cloud init conf.
This lead us to hardcode it first just for the sake of the demo but I believe it would be better to allow this to be passed and fetched from an environment parameter too similar to the secrets configs.

* [x] Run `make lint`
* [x] Run `make test`


MoeMahhouk:
I have ran the make tests too and all green besides one of the tests is failing.
But I checked on the develop branch and the same test is failing. The failing test seems not relevant with those changes

ZanCorDX:
This looks like a copy/paste from EnvOrInplaceValue, can you try to refactor it?
Ideally "env:" should be a constant.

ZanCorDX (approved):

metachris:
@MoeMahhouk could you rebase please?

crates/rbuilder/src/building/block_orders/share_bundle_merger.rs
```diff
@@ -214,12 +214,12 @@ impl<SinkType: SimulatedOrderSink> MultiBackrunManager<SinkType> {
...
 /// - Backruns: It contains a sub bundle with the user txs followed by backrunner sub bundles. (see [`ShareBundleMerger::break_down_backrun_bundle`] for a clearer definition)
-/// Merging example: SBundle(Bundle(tx)+br1) + SBundle(Bundle(tx)+br2) will become SBundle(Bundle(Bundle(tx)+br1),Bundle(Bundle(tx)+br2)).
-/// User Bundle may contain more that one tx (for the case approve + swap).
-/// This means what when we insert/remove orders (sbundles) we insert/remove the virtual bundles we generate and and not the original orders.
-/// We ONLY use this special flow for orders with some particular structure (see [`ShareBundleMerger::break_down_bundle`]).
-/// SBundle not complying with that will pass through and use the standard handling.
-/// @Pending evaluate if we can avoid to send changes no every order update and "flush" changes all together (since orders are usually processed on batches)
+///   Merging example: SBundle(Bundle(tx)+br1) + SBundle(Bundle(tx)+br2) will become SBundle(Bundle(Bundle(tx)+br1),Bundle(Bundle(tx)+br2)).
```
metachris, line 217: why adding these spaces?

MoeMahhouk: When I rebased and solved the conflicts. The lint pipeline blocked the merging due to those errors.
Here you can see them
https://github.com/flashbots/rbuilder/actions/runs/10141455345/job/28038732480

PS. interesting though that those files were not introduced by my PR but still blocked my PR to merge.
Also the make lint locally didn't show me to fix them. But I did them manually just to make the PR to progress forward

PR #67, opened 2024-07-16
Distribution logic

dvush:
## 📝 Summary

This implements a redistribution logic for the builder. It adds redistribution logic as part of the backtest and adds a binary to calculate output csv files.

## 💡 Motivation and Context

Redistribution refers to a process of dividing block profits between contribution signers.

To do that we
* built block and calculate its backtest profit
* exclude some signer and calculate profit as if this signer never sent any bundles
* the difference between two values (second one should be lower) is contribution by the given signer
* profit is divided proportionally to the contribution

For transactions sent through the rpc endpoint we can't use signer and instead we use tx origin or address specified as address to receive mev share refunds.

---

* [ ] Run `make lint`
* [ ] Run `make test`


crates/rbuilder/src/backtest/redistribute/cli/csv_output.rs
```diff
@@ -0,0 +1,52 @@
...
+    pub fn new(path: impl AsRef<Path>) -> io::Result<Self> {
+        let file = std::fs::OpenOptions::new()
+            .write(true)
+            .create(true)
+            .truncate(true)
+            .open(path)?;
+        Ok(Self { file })
+    }
+
+    pub fn write_header(&mut self) -> io::Result<()> {
```
ZanCorDX: Can write_header() be called inside new() and make it private?

crates/rbuilder/src/backtest/redistribute/cli/mod.rs
```diff
@@ -0,0 +1,134 @@
...
+    let cli = Cli::parse();
+
+    let config: Config = load_config_toml_and_env(cli.config)?;
+    config.base_config.setup_tracing_subsriber()?;
+
+    let mut historical_data_storage =
+        HistoricalDataStorage::new_from_path(&config.base_config.backtest_fetch_output_file)
+            .await?;
+
+    match cli.command {
```
ZanCorDX, line 62: Can we avoid the copy/paste on Block vs Range?

ZanCorDX:
Mempool txs are brought from Duration::minutes(3) ago (MempoolDumpsterDatasource) but in live buiding spawn_clean_orderpool_job keeps only TIME_TO_KEEP_BUNDLE_CANCELLATIONS: Duration = Duration::from_secs(60); are we handling this?

ferranbt:
I understand that the refund logic needs to be open source for the TEE builder and that part of that refund logic takes advantage of the bench tooling of the Rbuilder, but, could we move this logic to some TEE-builder specific repository?

This logic is part of a product offering from Flashbots and I do not think it belongs into the OSS version of the rbuilder.

ZanCorDX (changes requested):
Mentioned copy/paste and timing stuff.

crates/rbuilder/src/backtest/backtest_build_range.rs
```diff
@@ -399,6 +402,7 @@ impl CSVResultWriter {
 fn spawn_block_fetcher(
     mut historical_data_storage: HistoricalDataStorage,
     blocks: Vec<u64>,
+    ingored_signers: Vec<Address>,
```
ryanschneider: Typo: 
```suggestion
    ignored_signers: Vec<Address>,
```

crates/rbuilder/src/backtest/backtest_build_range.rs
```diff
@@ -408,13 +412,16 @@ fn spawn_block_fetcher(
...
-            let blocks = match historical_data_storage.read_blocks(blocks).await {
+            let mut blocks = match historical_data_storage.read_blocks(blocks).await {
                 Ok(res) => res,
                 Err(err) => {
                     warn!("Failed to read blocks from storage: {:?}", err);
                     return;
                 }
             };
+            for block in &mut blocks {
+                block.filter_out_ignored_signers(&ingored_signers);
```
ryanschneider:
```suggestion
                block.filter_out_ignored_signers(&ignored_signers);
```

dvush:
I found some issues here that needs to be addressed.

@ferranbt I think the core logic should be in this repo but maybe some wrapper can be in another. The motivation is to have this updated with the rest of the repo and ease of maintenance.

ferranbt:
> The motivation is to have this updated with the rest of the repo and ease of maintenance.

What is the maintenance burden? The bench framework?

dvush:
Diff is super wierd, its actually rebased. Closing and reopening

PR #69, opened 2024-07-17
Add playground integration

ferranbt:
## 📝 Summary

This PR implements and integration test using the [builder-playground](https://github.com/flashbots/builder-playground):

- It creates a playground.rs utility library to interact with the playground. **In this first iteration, most of the variables (e.g. private keys, urls, etc...) are hardcoded**.
- It creates a simple integration test that ensures that a transaction sent to the EL gets mined by the rbuilder and included on the block through mev-boost-relay.
- It runs the integration test in CI after the unit tests have passed.

The integration test works directly with the `rbuilder` compiled binary.

## 💡 Motivation and Context

---

* [ ] Run `make lint`
* [ ] Run `make test`


.github/workflows/checks.yaml
```diff
@@ -58,5 +54,48 @@ jobs:
...
+      - name: Run integration tests
+        run: cargo test --package rbuilder --lib -- integration
+        env:
+          PLAYGROUND_DIR: /tmp/playground
+
+      - name: Move test integration logs to playground logs to archive
+        # This steps fails if the test fails early and the playground logs dir has not been created
+        if: ${{ failure() }}
+        run: |
+          mv integration_logs /tmp/playground/logs
```
metachris, line 93: is this step needed? couldn't you directly upload the folder as artifact without moving?

rbuilder.toml
```diff
@@ -0,0 +1,49 @@
+log_json = false
+log_level = "info,rbuilder=debug"
+telemetry_port = 6060
+telemetry_ip = "0.0.0.0"
+
+chain = "/Users/ferranbt/go/src/github.com/flashbots/builder-playground/local-testnet/genesis.json"
```
metachris: would be ideal if we could put the local chain data into `~/.builder-playground/...` where the other builder playground stuff is located in. that way there could be a single example config that works for everyone.

metachris (approved):

ZanCorDX (approved):

metachris (approved):

Issue #79, opened 2024-07-23
Multiple RUSTSEC security advisories across several dependencies


jmcph4:
As of `ae24579813ba4eda3d5c5469f4a107900516d7ee`, there are multiple Rust security advisories affecting the codebase. Specifically,

- [RUSTSEC-2024-0344](https://rustsec.org/advisories/RUSTSEC-2024-0344)
- [RUSTSEC-2024-0332](https://rustsec.org/advisories/RUSTSEC-2024-0332)
- [RUSTSEC-2024-0336](https://rustsec.org/advisories/RUSTSEC-2024-0336)
- [RUSTSEC-2024-0347](https://rustsec.org/advisories/RUSTSEC-2024-0347)
- [RUSTSEC-2024-0346](https://rustsec.org/advisories/RUSTSEC-2024-0346)
- [RUSTSEC-2024-0320](https://rustsec.org/advisories/RUSTSEC-2024-0320)
- [RUSTSEC-2023-0055](https://rustsec.org/advisories/RUSTSEC-2023-0055)

Most of these are via Reth dependencies and would thus be solved by the existing Reth-related tickets (for instance, #51). The final one is via the Rust bindings for mempool-dumpster and would be solved by dvush/mempool-dumpster-rs#1.

It would also likely be appropriate to add a `cargo-audit` run to the existing CI pipeline.

Output of `cargo-audit` follows:

```
$ git rev-parse HEAD
ae24579813ba4eda3d5c5469f4a107900516d7ee
$ cargo audit
Crate:     curve25519-dalek
Version:   4.1.2
Title:     Timing variability in `curve25519-dalek`'s `Scalar29::sub`/`Scalar52::sub`
Date:      2024-06-18
ID:        RUSTSEC-2024-0344
URL:       https://rustsec.org/advisories/RUSTSEC-2024-0344
Solution:  Upgrade to >=4.1.3
Dependency tree:
curve25519-dalek 4.1.2
└── ed25519-dalek 2.1.1
    ├── libp2p-identity 0.2.8
    │   ├── reth-discv5 0.2.0-beta.6
    │   │   └── reth-network 0.2.0-beta.6
    │   │       ├── reth-node-events 0.2.0-beta.6
    │   │       │   ├── reth-node-builder 0.2.0-beta.6
    │   │       │   │   ├── reth-node-ethereum 0.2.0-beta.6
    │   │       │   │   │   └── reth 0.2.0-beta.6
[...]

PR #51, opened 2024-07-10
chore(deps): bump reth to v1.0.3

chirag-bgh:
## 📝 Summary

Update reth to latest v1.0.3 #47 

## 💡 Motivation and Context

---

## ✅ I have completed the following steps:

* [x] Run `make lint`
* [ ] Run `make test`
* [ ] Added tests (if applicable)

PR #80, opened 2024-07-23
Add audit job to CI pipeline

jmcph4:
## 📝 Summary

 - Add `cargo audit` invocation to `Makefile` as `make audit`
 - Add `make audit` to existing CI pipeline

## 💡 Motivation and Context

See #79 for context.

---

* [x] Run `make lint`
* [x] Run `make test`
* [x] ~~Added tests (if applicable)~~

### DIFF
--- a/.github/workflows/checks.yaml
+++ b/.github/workflows/checks.yaml
@@ -57,6 +57,7 @@ jobs:
       - name: Install native dependencies
         run: sudo apt-get install -y libsqlite3-dev
 
-      # lint and test
+      # lint, test, and audit
       - run: make lint
       - run: make test
+      - run: make audit
--- a/Makefile
+++ b/Makefile
@@ -49,6 +49,10 @@ fmt: ## Format the code
 	cargo fix --allow-staged
 	cargo clippy --fix --allow-staged
 
+.PHONY: audit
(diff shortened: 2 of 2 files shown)


dvush (approved):

metachris:
This is great, thanks! 

Merging this as-is would make the checks CI step fail. Are the audit issues easy to resolve?

jmcph4:
> Merging this as-is would make the checks CI step fail.

Yeah I knew this would break CI but wanted to file a PR regardless; I agree that it shouldn't be merged in prior to actual fixes.

> Are the audit issues easy to resolve?

It's essentially blocked by two things:

 - Updating our Reth dependencies (#51)
 - Merging in dvush/mempool-dumpster-rs#2, which will fix the same issues upstream. Note that after this occurs, we can either
   - Publish a new crate version for `mempool-dumpster-rs`, or
   - Change the dependency in `rbuilder` to point to the Git repo directly

Personally, I'm leaning towards the latter resolution.

metachris:
After some debate, we prefer not to merge this because it will lead to lots of false positives and block merging PRs.

Thanks for the contribution, looking forward to your next one <3

PR #82, opened 2024-07-24
Fix consistency check for merged sbundles

dvush:
There were situations where we had merged sbundles that looked like this

bundle1:
   user_tx, allow_revert: no
   backrun1

bundle 2:
  user_tx: allow_revert yes
  backrun

merged sbundle is concat of these two

and consistency check would think that user_tx wat not allowed to revert enven though there is a situation where we include bundle2


ryanschneider (approved):
lgtm!

PR #83, opened 2024-07-24
[wip - don't merge] initial bundle sponsorship code

### LINKED ISSUE #25: feat: bundle sponsorship
Some transactions fail because they lack funds, but they generate mev and are actually profitable to execute if they did have funds to do so. We could "sponsor" them as a builder by transferring some eth to the account that lacks funds before its transaction is executed.

bertmiller:
## 📝 Summary
This work starts to implement #25, bundle sponsorship. It shows how you can catch the InvalidTransaction::LackOfFundForMaxFee error and parse it to grab the necessary information to attempt a bundle sponsorship.

**I started this PR months ago but didn't carry it out because at the time reth had a bug and was returning the wrong information. I am no longer actively working on it as my priorities have changed, so I'd encourage someone else to pick it up from here.**

The next steps for this work:
* Simulate sponsored bundles
* Evaluate if they are profitable to sponsor and handle the results of simulation
* Alter `SimulatedOrder` to include info about sponsorship if desirable
* Parse and handle sponsored bundles when filling orders (e.g. by inserting a sponsor tx with the right nonce)

But I haven't thought it out thoroughly!

## 💡 Motivation and Context
See #25, we might want this because it allows us to execute some transactions and make fees from them that we otherwise wouldn't be able to - thus leading to more valuable blocks.

---

* [ ] Run `make lint`
* [ ] Run `make test`

### DIFF
--- a/crates/rbuilder/src/building/order_commit.rs
+++ b/crates/rbuilder/src/building/order_commit.rs
@@ -174,7 +174,7 @@ pub struct TransactionOk {
     pub receipt: Receipt,
 }
 
-#[derive(Error, Debug, Eq, PartialEq)]
+#[derive(Clone, Error, Debug, Eq, PartialEq)]
 pub enum TransactionErr {
     #[error("Invalid transaction: {0:?}")]
     InvalidTransaction(InvalidTransaction),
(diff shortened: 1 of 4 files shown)


chirag-bgh:
Hey, I'm interested in implementing this feature. I would take this up from here.

PR #98, opened 2024-08-05
chore: cleanup example config: comment internal blocks_processor_url

metachris:
## 📝 Summary

Remove internal URL from example config to reduce confusion

---

* [x] Run `make lint`
* [x] Run `make test`

### DIFF
--- a/config-live-example.toml
+++ b/config-live-example.toml
@@ -21,7 +21,7 @@ blocklist_file_path = "./blocklist.json"
 dry_run = true
 dry_run_validation_url = "http://localhost:8545"
 
-blocks_processor_url = "http://block_processor.internal"
+blocks_processor_url = ""
 ignore_cancellable_orders = true
 
 sbundle_mergeabe_signers = []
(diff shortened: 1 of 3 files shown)


config-live-example.toml
```diff
@@ -21,7 +21,7 @@ blocklist_file_path = "./blocklist.json"
 dry_run = true
 dry_run_validation_url = "http://localhost:8545"
 
-blocks_processor_url = "http://block_processor.internal"
```
ferranbt, line 24: Did you try to run this on the playground? I think the value is not optional and if it is empty it is going to fail because the url is malformed (e.g. empty).

ZanCorDX: Agree, it's Option<> so we just have to remove the line (or comment and add another comment explaining)

ZanCorDX (changes requested):

metachris:
Thanks for the comments. Makes sense, just commented out the line and left a comment about it. Tested with builder-playground.

config-live-example.toml
metachris: fixed

ferranbt (approved):

ZanCorDX (approved):

PR #102, opened 2024-08-08
Doc explaining reorg losses.

ZanCorDX:
## 📝 Summary

Doc explaining reorg losses.

* [ ] Run `make lint`
* [ ] Run `make test`


REORG_LOSSES.md
```diff
@@ -0,0 +1,19 @@
+# Reorg losses
```
metachris: Could you create a `/docs/` folder and add this there? (imo ideally with a lowercase filename)

I think it's better to start the directory because over time we'll accumulate more docs and it's better to have them together in a single place, instead of scattered on the project root. Alternatively, the wiki might also be a good place, but that has the downside of then not being self-contained in a clone of the repo.

REORG_LOSSES.md
```diff
@@ -0,0 +1,19 @@
...
+
+From time to time ethereum network forks for a short time causing block reorgs.
```
metachris: Please grammar/typo check.

Also might add a link to https://etherscan.io/blocks_forked

README.md
```diff
@@ -16,6 +16,8 @@ It is designed to provide a delightful developer experience, enabling community
 
 ## Running rbuilder
 
+**Warning**: Even if they are rare, before running a builder you should be aware of [`reorg losses`](./REORG_LOSSES.md).
```
metachris: I would not add this as the top info on "running rbuilder". 

It seems better to place this line at the _end_ of the "Live" subsection.

REORG_LOSSES.md
```diff
@@ -0,0 +1,19 @@
...
+When a reorg happens all the txs from the losing fork goes to the mempool. This includes any private transaction, particularly any builder generated tx like the one we add at the end of the block to pay to the validator.
```
jakubhruby7: in docs like this one let's use "transaction" rather than "tx" for clarity

REORG_LOSSES.md
```diff
@@ -0,0 +1,19 @@
...
+
+If this happens, the builder ends up paying the bid to a random validator without winning any block! Notice that **this happens even if this tx gives not tip to the block**, probably because some builders are such good guys that when all paying transactions are included in the block they are building (and some gas is left) they include free txs.
```
jakubhruby7: "good guys" -> to casual language for a doc like this imo

jakubhruby7 (changes requested):
Outside of my comments, let's also fix grammar. It's incorrect in a few places in different parts of the doc.

metachris (approved):

jakubhruby7 (approved):

PR #118, opened 2024-08-13
Decouple l1 config from builder config

ferranbt:
## 📝 Summary

This PR moves the L1 specific config from BaseConfig (i.e. relay url, beacon chain url, etc...) into a new `L1Config` struct that gets flattened into the main `Config`. This PR does not introduce any logic change. Now, `BaseConfig` has generic types for both block sink and slot source. 

* [x] Run `make lint`
* [x] Run `make test`

### DIFF
--- a/crates/rbuilder/src/live_builder/config.rs
+++ b/crates/rbuilder/src/live_builder/config.rs
@@ -1,6 +1,9 @@
 //! Config should always be deserializable, default values should be used
 //!
+//!
+use super::{base_config::BaseConfig, building::relay_submit::RelaySubmitSinkFactory};
 use crate::{
+    beacon_api_client::Client,
     building::{
         builders::{
             ordering_builder::{OrderingBuilderConfig, OrderingBuildingAlgorithm},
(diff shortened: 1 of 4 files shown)


crates/rbuilder/src/live_builder/config.rs
```diff
@@ -68,7 +244,38 @@ impl LiveBuilderConfig for Config {
...
+
+        let relays = self.l1_config.relays()?;
+        let sink_factory = RelaySubmitSinkFactory::new(submission_config, relays.clone());
+
+        let payload_event = MevBoostSlotDataGenerator::new(
+            self.l1_config.beacon_clients()?,
+            relays,
+            self.base_config.blocklist()?,
+            cancellation_token.clone(),
+        );
```
ZanCorDX, line 273: Shouldn't this be inside L1Config?

ferranbt: Not sure about that, this is a function that returns the concrete implementation of the LiveBuilder so it needs to have a reference both to the l1 config and the builder config.

ZanCorDX (approved):

PR #135, opened 2024-08-20
Stream unsealed blocks

ZanCorDX:
## 📝 Summary

Introduced **BlockBuildingHelper**: A new refactoring structure that contains all the necessary state to build blocks. 

This pull request modifies **BlockBuildingAlgorithms** to output a **BlockBuildingHelper** instead of a **Block**, removing the finishing steps (payout transaction and root hash) from the algorithms.

A new layer has been introduced to transition from **BlockBuildingHelper** to **Block**, followed by the relay submission, specifically tailored for L1. This layer consists of the BlockFinisher, which communicates with the bidder to determine the next steps and, when necessary, propagates the block.

This change is nearly 100% backward compatible; the sequence of operations remains the same, although they are now executed by different objects.

## 💡 Motivation and Context

- Allow best bidding strategies (rebidding)
- Improve L2 compatibility (bidding is not part of the core anymore).
- Allow bottom of block building (a **BlockBuildingHelper** could be upgraded). 

---

* [x] Run `make lint`
* [x] Run `make test`

### DIFF
--- a/crates/rbuilder/src/bin/dummy-builder.rs
+++ b/crates/rbuilder/src/bin/dummy-builder.rs
@@ -5,23 +5,22 @@
 //! This is NOT intended to be run in production so it has no nice configuration, poor error checking and some hardcoded values.
 use std::{path::PathBuf, sync::Arc, thread::sleep, time::Duration};
 
-use alloy_primitives::U256;
 use jsonrpsee::RpcModule;
 use rbuilder::{
     beacon_api_client::Client,
     building::{
         builders::{
-            finalize_block_execution, Block, BlockBuildingAlgorithm, BlockBuildingAlgorithmInput,
-            BlockBuildingSink, BuilderSinkFactory, OrderConsumer,
+            block_building_helper::{BlockBuildingHelper, BlockBuildingHelperFromDB},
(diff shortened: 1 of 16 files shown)


crates/rbuilder/src/bin/dummy-builder.rs
```diff
@@ -117,31 +114,33 @@ async fn main() -> eyre::Result<()> {
...
+impl UnfinishedBlockBuildingSink for TracingBlockSink {
+    fn new_block(&self, block: Box<dyn BlockBuildingHelper>) {
         info!(
-            tx_count =? block.sealed_block.body.len(),
+            order_count =? block.built_block_trace().included_orders.len(),
             "Block generated. Throwing it away!"
         );
     }
+
+    fn can_use_suggested_fee_recipient_as_coinbase(&self) -> bool {
```
ferranbt, line 141: Does this need to be part of the sink? Could it be part of the factory?

ferranbt (approved):

crates/rbuilder/src/bin/dummy-builder.rs
ZanCorDX: It's needed on a block-by-block basis. Even worst, it changes during the block.

PR #139, opened 2024-08-21
Redistribution output rich data + approx. identity rule

dvush:
## 📝 Summary

* output more internal data from redistribution step
* tweak rule to be more in line with the doc 

## 💡 Motivation and Context

---

* [ ] Run `make lint`
* [ ] Run `make test`


ZanCorDX (approved):
Should we change backtest DB version?

dvush:
There is a  change to the result store db and since the values are serialized with serde default this should be compatible with previous version

ryanschneider (approved):
lgtm!

PR #141, opened 2024-08-22
Make the block builder pool a block building algorithm

ferranbt:
## 📝 Summary

Right now there is an ad-hoc `BlockBuilderPool` entity that lives inside the `live_builder` and manages a list of `BlockBuildingAlgorithm` trait implementations. The pool takes the input to build a block and spawns different algorithms with the same sink destination.

The `live_builder` uses a concrete reference of the `BlockBuilderPool` and initialises it with a list of builders.

This PR converts the `BlockBuilderPool` in an implementation of the `BlockBuildingAlgorithm` trait. Then, the `live_builder` does not need to have a concrete reference of the `BlockBuilderPool` but just a `dyn BlockBuildingAlgorithm` reference which can be a single block building algorithm or a pool of them.

## 💡 Motivation and Context

---

* [x] Run `make lint`
* [x] Run `make test`

### DIFF
--- a/crates/rbuilder/src/live_builder/mod.rs
+++ b/crates/rbuilder/src/live_builder/mod.rs
@@ -10,11 +10,17 @@ mod watchdog;
 
 use crate::{
     building::{
-        builders::{BlockBuildingAlgorithm, UnfinishedBlockBuildingSinkFactory},
+        builders::{
+            BlockBuildingAlgorithm, BlockBuildingAlgorithmInput, UnfinishedBlockBuildingSinkFactory,
+        },
         BlockBuildingContext,
     },
     live_builder::{
-        order_input::{start_orderpool_jobs, OrderInputConfig},
+        building::BlockBuildingPool,
+        order_input::{
+            order_replacement_manager::OrderReplacementManager, orderpool::OrdersForBlock,
+            start_orderpool_jobs, OrderInputConfig,
+        },
         simulation::OrderSimulationPool,
         watchdog::spawn_watchdog_thread,
     },
@@ -23,7 +29,6 @@ use crate::{
 };
 use ahash::HashSet;
 use alloy_primitives::{Address, B256};
-use building::BlockBuildingPool;
 use eyre::Context;
 use jsonrpsee::RpcModule;
 use payload_events::MevBoostSlotData;
@@ -35,9 +40,12 @@ use reth_chainspec::ChainSpec;
 use reth_db::database::Database;
 use std::{cmp::min, path::PathBuf, sync::Arc, time::Duration};
 use time::OffsetDateTime;
-use tokio::{sync::mpsc, task::spawn_blocking};
+use tokio::{
+    sync::{broadcast, mpsc},
+    task::spawn_blocking,
+};
 use tokio_util::sync::CancellationToken;
-use tracing::{debug, error, info, warn};
+use tracing::{debug, error, info, trace, warn};
 
 /// Time the proposer have to propose a block from the beginning of the slot (https://www.paradigm.xyz/2023/04/mev-boost-ethereum-consensus Slot anatomy)
 const SLOT_PROPOSAL_DURATION: std::time::Duration = Duration::from_secs(4);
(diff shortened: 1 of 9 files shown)


crates/rbuilder/src/live_builder/mod.rs
```diff
@@ -123,15 +134,18 @@ impl<DB: Database + Clone + 'static, BuilderSourceType: SlotSource>
...
 
+        /*
         let mut builder_pool = BlockBuildingPool::new(
             self.provider_factory.clone(),
             self.builders,
             self.sink_factory,
             orderpool_subscriber,
             order_simulation_pool,
         );
+        */
```
ZanCorDX: remove!!

crates/rbuilder/src/live_builder/mod.rs
```diff
@@ -74,7 +82,7 @@ pub struct LiveBuilder<DB, BlocksSourceType: SlotSource> {
     pub global_cancellation: CancellationToken,
 
     pub sink_factory: Box<dyn UnfinishedBlockBuildingSinkFactory>,
-    pub builders: Vec<Arc<dyn BlockBuildingAlgorithm<DB>>>,
+    pub builder: Option<Arc<dyn BlockBuildingAlgorithm<DB>>>, // doing the Option because there is a fuunction that creates the live_builder without the builder.
```
ZanCorDX: Can we remove the Option and go for Null object pattern?

crates/rbuilder/src/live_builder/mod.rs
```diff
@@ -215,12 +229,49 @@ impl<DB: Database + Clone + 'static, BuilderSourceType: SlotSource>
...
+                    ctx: block_ctx,
+                    sink: builder_sink,
+                    input: broadcast_input.subscribe(),
+                    cancel: block_cancellation,
+                };
+
+                tokio::spawn(multiplex_job(simulations_for_block.orders, broadcast_input));
+
+                if let Some(builder) = self.builder.as_ref() {
+                    builder.build_blocks(input);
```
ZanCorDX: Even though you know that BlockBuildingPool is not blocking, builders are expected to be blocking. We should tokio::task::spawn_blocking this.

crates/rbuilder/src/live_builder/mod.rs
```diff
@@ -215,12 +229,49 @@ impl<DB: Database + Clone + 'static, BuilderSourceType: SlotSource>
...
+                }
+            }
```
ZanCorDX: Can you put this into a func so the code is more readable like before?

crates/rbuilder/src/live_builder/mod.rs
ferranbt: Sure, can you point me how to do that?

crates/rbuilder/src/live_builder/mod.rs
ferranbt: Which one do you want as spawn_blocking?

ZanCorDX (changes requested):
Due to type restrictions we didn't have time to polish (BlockBuildingAlgorithmInput having a very specific communication method: broadcast::Receiver) you were forced to add an unnecesary background task on LiveBuilder::run():
```
tokio::spawn(multiplex_job(simulations_for_block.orders, broadcast_input));
```
I think this can be avoided by using a trait since the only functions called are try_recv and recv, both implemented on broadcast::Receiver and mpsc::Receiver.
You could either add wrapper objects or extend mpsc::Receiver/broadcast/Receiver (this is cheaper).

ferranbt:
> You could either add wrapper objects or extend mpsc::Receiver/broadcast/Receiver (this is cheaper)

Any hints on how to do this?

crates/rbuilder/src/live_builder/mod.rs
ZanCorDX: builder.build_blocks(input);

ferranbt: Check again, I think this is resolved now.

ferranbt:
Shot out to @liamaharon for the help on the fixes.

crates/rbuilder/src/live_builder/mod.rs
```diff
@@ -74,7 +79,7 @@ pub struct LiveBuilder<DB, BlocksSourceType: SlotSource> {
     pub global_cancellation: CancellationToken,
 
     pub sink_factory: Box<dyn UnfinishedBlockBuildingSinkFactory>,
-    pub builders: Vec<Arc<dyn BlockBuildingAlgorithm<DB>>>,
+    pub builder: Arc<dyn BlockBuildingAlgorithm<DB>>, // doing the Option because there is a fuunction that creates the live_builder without the builder.
```
liamaharon, line 82: Any reason dyn was chosen over making the builder algorithm generic?

ZanCorDX: Generic types explosion (harder to maintain).

PR #143, opened 2024-08-25
Make e2e test more deterministic

ferranbt:
## 📝 Summary

This PR does not fix the e2e test but makes it a bit more deterministic to execute. I feel the problem is that due to https://github.com/flashbots/builder-playground/issues/14 issue, the `rbuilder` it not consistent in building blocks at the correct time for the playground. So, it misses building some blocks.

The previous e2e test would send the transaction to the EL node (to be picked up by the builder) and the acceptance condition was for the transaction to be mined and that block to be mined by the local builder. However, since the builder was not consistent, the transaction would be mined by the local EL node instead and trigger the acceptance check.

In this fix, I am only sending the transaction to the builder, so that the EL node cannot include it. Then, if the transaction gets mined, it was for sure done by the builder. I have tested it locally multiple times (>10) and it only failed once and I think due to https://github.com/flashbots/builder-playground/issues/14.

## 💡 Motivation and Context

---

* [x] Run `make lint`
* [x] Run `make test`

### DIFF
--- a/crates/rbuilder/src/integration/playground.rs
+++ b/crates/rbuilder/src/integration/playground.rs
@@ -108,6 +108,10 @@ impl Playground {
         Ok(event)
     }
 
+    pub fn rbuilder_rpc_url(&self) -> &str {
+        "http://localhost:8645"
+    }
+
     pub fn el_url(&self) -> &str {
         "http://localhost:8545"
     }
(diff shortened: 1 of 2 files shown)


ryanschneider (approved):
lgtm!

PR #152, opened 2024-08-26
Redistribution fixes

dvush:
## 📝 Summary

* now for individual order exclusion data we also exclude orders that are "similar" to it from the same identity
* now when profit is changed we also consider this a conflict
* user first (before it was last) address for the share bundle identity processed by mev share node

## 💡 Motivation and Context

---

* [ ] Run `make lint`
* [ ] Run `make test`


ryanschneider (approved):
lgtm!

PR #154, opened 2024-08-26
Bug fix for failed cancelled sinumations.

ZanCorDX:
## 📝 Summary

If a simulation fails this error is not handled.
If after that, a cancelation comes for the failed simulation we propagate it downstream generating a cancellation for an order we never sent.

## 💡 Motivation and Context

We have an ERROR log "remove_order for not inserted order". This fix could solve it.

---

* [x] Run `make lint`
* [x] Run `make test`


ryanschneider (approved):
lgtm

ferranbt (approved):
It would be great if we can add a unit test.

ZanCorDX:
> It would be great if we can add a unit test.

Yes but sadly it's not that easy due to threading stuff (how do we know that the simulation already failed?).

PR #156, opened 2024-08-27
Extend datasource trait

dvush:
## 📝 Summary

Slightly change DataSource trait to make data source implementation more flexible. 

## 💡 Motivation and Context

---

* [ ] Run `make lint`
* [ ] Run `make test`

### DIFF
--- a/crates/rbuilder/src/backtest/fetch/data_source.rs
+++ b/crates/rbuilder/src/backtest/fetch/data_source.rs
@@ -2,14 +2,18 @@ use crate::backtest::{BuiltBlockData, OrdersWithTimestamp};
 use alloy_primitives::B256;
 use async_trait::async_trait;
 
+#[derive(Debug, Clone)]
+pub struct DatasourceData {
+    pub orders: Vec<OrdersWithTimestamp>,
+    pub built_block_data: Option<BuiltBlockData>,
+}
+
 /// DataSource trait
 ///
 /// This trait is used to fetch data from a datasource
 #[async_trait]
 pub trait DataSource: std::fmt::Debug {
-    async fn get_orders(&self, block: BlockRef) -> eyre::Result<Vec<OrdersWithTimestamp>>;
-
-    async fn get_built_block_data(&self, block_hash: B256) -> eyre::Result<Option<BuiltBlockData>>;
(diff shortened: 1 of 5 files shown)


crates/rbuilder/src/backtest/fetch/data_source.rs
```diff
@@ -20,19 +24,21 @@ impl Clone for Box<dyn DataSource> {
...
+    pub landed_block_hash: Option<B256>,
 }
 
 impl BlockRef {
-    pub fn new(block_number: u64, block_timestamp: u64) -> Self {
+    pub fn new(block_number: u64, block_timestamp: u64, landed_block_hash: Option<B256>) -> Self {
         Self {
             block_number,
             block_timestamp,
+            landed_block_hash,
```
ferranbt, line 41: Is this the block_hash or something different?

ferranbt (approved):

crates/rbuilder/src/backtest/fetch/data_source.rs
```diff
@@ -2,14 +2,18 @@ use crate::backtest::{BuiltBlockData, OrdersWithTimestamp};
 use alloy_primitives::B256;
 use async_trait::async_trait;
 
+#[derive(Debug, Clone)]
+pub struct DatasourceData {
+    pub orders: Vec<OrdersWithTimestamp>,
+    pub built_block_data: Option<BuiltBlockData>,
```
ferranbt, line 8: Eventually, I think it would make sense to move built_block_data to a different trait since there should be only one data source that returns block data.

crates/rbuilder/src/backtest/fetch/data_source.rs
dvush: Yes, its a block hash of the landed block

crates/rbuilder/src/backtest/fetch/data_source.rs
dvush: It can be just a function(or trait with one funciton) that you pass into the fetcher that will give you all these things at once. The default implementation will use mempool dumpster + optional flashbots db as part of one struct to implement this function.

PR #159, opened 2024-08-27
Comment crates/rbuilder/src/building/evm_inspector.rs

dvush:
## 📝 Summary

Comment crates/rbuilder/src/building/evm_inspector.rs

## 💡 Motivation and Context

---

* [ ] Run `make lint`
* [ ] Run `make test`

### DIFF
--- a/crates/rbuilder/src/building/evm_inspector.rs
+++ b/crates/rbuilder/src/building/evm_inspector.rs
@@ -36,6 +36,9 @@ impl<'a> UsedStateEVMInspector<'a> {
         }
     }
 
+    /// This method is used to mark nonce change as a slot read / write.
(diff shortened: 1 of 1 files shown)


ZanCorDX (approved):

ferranbt (approved):

jmcph4:
This is a very useful comment but there could be a lot more explanation in this module especially. Mainly,

 - Why `B256` and not `U256` when interacting with EVM words (this is because the word being retrieved represents a storage slot, which is a digest)
 - Why `sload_tmp_key` exists (there's already a comment here but it should be expanded upon in my opinion)
 - Why `RBuilderEVMInspector` composes two inspectors together and why they're individually useful
 - Why the ordering of read and write slots is important

I was going to open a PR doing this due to questions I had in Telegram yesterday but obviously @dvush beat me to the punch.

PR #160, opened 2024-08-28
Wait for rbuilder running before running integration test

ferranbt:
## 📝 Summary

The new integration test introduced in #143 requires the rbuilder to be running to work. However, the playground setup was not waiting for it. This PR fixes that. It adds a loop after the rbuilder creation that checks in the output log for a specific log "RPC server job: started" that signals that the builder is running and the test can start.

## 💡 Motivation and Context

---

* [x] Run `make lint`
* [x] Run `make test`

### DIFF
--- a/crates/rbuilder/src/integration/playground.rs
+++ b/crates/rbuilder/src/integration/playground.rs
@@ -15,7 +15,8 @@ use std::{
     path::PathBuf,
     process::{Child, Command},
     str::FromStr,
-    time::SystemTime,
+    thread,
+    time::{Instant, SystemTime},
 };
 use time::{format_description, OffsetDateTime};
 use url::Url;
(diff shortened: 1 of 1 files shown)


metachris (approved):
thxxx

crates/rbuilder/src/integration/playground.rs
```diff
@@ -94,6 +96,26 @@ impl Playground {
...
+            let mut contents = String::new();
+            file.read_to_string(&mut contents)
+                .map_err(|_| PlaygroundError::SetupError)?;
+
+            if contents.contains("RPC server job: started") {
+                break;
+            }
+
+            thread::sleep(std::time::Duration::from_millis(100));
+        }
```
liamaharon, line 117: Something like this would be more idiomatic and avoid spamming io ops https://github.com/jmagnuson/linemux. I wouldn't call it a "must have" though.

liamaharon (approved):

crates/rbuilder/src/integration/playground.rs
ferranbt: Yeah, I was hoping to do an http request to check some health check endpoint. But could not make it work.

Issue #58, opened 2024-07-13
feat: improve UsedStateEVMInspector by adding other op-codes that read or write


bertmiller:
During simulation we use an EVM inspector to get state that is used by a transaction by looking at the storage slots that a transaction reads and writes from. We could supplement this by adding other op-codes that read or write state. A few off the top of my head that we should add:
* balance
* selfbalance
* create (balance and nonce is read)
* create2 (balance and nonce is read)
* selfdestruct (balance and nonce is read)

Do we need to support EXTCODECOPY, EXTCODEHASH, or EXTCODESIZE too? Not sure but I think so because some transactions might condition their execution on these and others might deploy contracts.

The goal would be to improve our coverage of what state is being used and how.

https://github.com/flashbots/rbuilder/blob/develop/crates/rbuilder/src/building/evm_inspector.rs

nikoseven:
I'd like to work on this.
I'm considering adding three more fields to UsedStateTrace. Does that sound good?

```
pub struct UsedStateTrace {
     ...
    // account read set
    pub balance_read_set: HashMap<Address, U256>,
    // list for contract create/destruct
    pub created_contract_addr_list: Vec<Address>,
    pub destructed_contract_addr_list: Vec<Address>,
}
```

ZanCorDX:
To have all the balance information we might need I believe it would be better to have something like this:

```
   // first balance account read
    pub balance_read_set: HashMap<Address, U256>,
    // last balance account written
    pub balance_write_set: HashMap<Address, U256>,
```
What do you think?

ZanCorDX:
The names we already had were not the best, read_set/write_set are not very meaningful.
It might be nice to correct them (if possible in another PR) to something like read_slot_values/written_slot_values (or read_slots/written_slots?) so your fields can be more like read_balances/written_balances and created_contracts/destructed_contracts.
What do you think?

PR #161, opened 2024-08-28
feat: improve UsedStateEVMInspector by adding more op-codes

nikoseven:
## 📝 Summary
Added support for SELFBALANCE, BALANCE, CALL, CREATE, and SELFDESTRUCT op-codes in UsedStateEVMInspector.
New test cases for evm inspector have been added.

## 💡 Motivation and Context
Related discussion in https://github.com/flashbots/rbuilder/issues/58.
For opcodes like EXTCODECOPY, EXTCODEHASH, and EXTCODESIZE, I thought we could add them in the future with a concrete use case, so that we can properly design the datastruct for it.

---

* [x] Run `make lint`
* [x] Run `make test`


dvush:
@nikoseven nice, thank you. I'll look into it soon.

ZanCorDX:
Hi!
We are considering adding some benchmarks before adding some extra weight to the execution code.
In the meantime, I see the word "Ephernal", did you mean "Ephemeral"?

nikoseven:
> Hi! We are considering adding some benchmarks before adding some extra weight to the execution code. In the meantime, I see the word "Ephernal", did you mean "Ephemeral"?

Fixed the typo. Thanks!
I'll wait for updates after benchmark being added.

dvush (approved):
I don't see any issue. It looks great. One thing though, we updated reth and revm so can you please rebase. 

re additonal allocations / performance: I don't think that this changes the pictrure in any way
1. we were already instrumenting step
2. this adds call/call_end instrumentation that runs very rarely
3. we were already allocating hashmap for each slot and here it allocates a tiny bit of data that is much smaller than that (the order is just the number of internal calls)

crates/rbuilder/src/building/evm_inspector.rs
```diff
@@ -104,9 +128,65 @@ where
...
+        outcome: revm::interpreter::CreateOutcome,
+    ) -> revm::interpreter::CreateOutcome {
+        if let Some(addr) = outcome.address {
+            self.used_state_trace.created_contracts.push(addr);
+        }
+        outcome
+    }
+
+    fn selfdestruct(&mut self, contract: Address, target: Address, value: U256) {
+        self.used_state_trace.destructed_contracts.push(contract);
```
dvush, line 176: This should check if contract was already destroyed and return if yes. Selfdestruct can be called multiple times in one tx.

crates/rbuilder/src/building/evm_inspector.rs
```diff
@@ -104,9 +128,65 @@ where
...
+            }
             _ => (),
         }
     }
+
+    fn call(&mut self, _: &mut EvmContext<DB>, inputs: &mut CallInputs) -> Option<CallOutcome> {
+        if let Some(transfer_value) = inputs.transfer_value() {
+            if !transfer_value.is_zero() {
+                *self
+                    .used_state_trace
```
dvush, line 149: This would be incorrect if call fails.

dvush (approved):
There is a small issue that would make balance part of the trace incorrect. I don't think that it must be fixed for this to be merged because our slot logic was not handling reverts anyway.

PR #164, opened 2024-09-03
`redacted` and `full` telemetry servers

liamaharon:
- ~Creates a new module `server` inside `live_builder`~
  - ~Server exposes a single fn `start`, which starts an actix server on the given port~
  - ~Initial server implementation just exposes one route `/health` that responds 200~
- ~Adds new base config item `live_builder_server_port`~
- ~Server is spawned when running the `live_builder` cli~

## Background 

- TDX builders cannot expose the existing telemetry server port, due to concerns about sensitive operational information leakage. This information may require a time-delay or additional sanitisation before being exposed outside of the secure machine.
- TDX builders require some other means to know if the builder is healthy or not.

## Solution

- Separated the telemetry server into two servers: `full` and `redacted`. 
  - `full`: what we have now, exposes all operational data without regard for whether it is sensitive
  - `redacted`: new server that initially only exposes a healthcheck endpoint (can be extended later). This server is safe to always be exposed by tdx builders.

### DIFF
--- a/crates/rbuilder/src/live_builder/base_config.rs
+++ b/crates/rbuilder/src/live_builder/base_config.rs
@@ -43,10 +43,10 @@ const ENV_PREFIX: &str = "env:";
 #[derive(Debug, Clone, Deserialize, PartialEq, Eq)]
 #[serde(default, deny_unknown_fields)]
 pub struct BaseConfig {
-    pub debug_server_port: u16,
-    pub debug_server_ip: Option<String>,
-    pub opaque_server_port: u16,
-    pub opaque_server_ip: Option<String>,
+    pub full_telemetry_server_port: u16,
+    pub full_telemetry_server_ip: Option<String>,
+    pub restricted_telemetry_server_port: u16,
+    pub restricted_telemetry_server_ip: Option<String>,
     pub log_json: bool,
     log_level: EnvOrValue<String>,
     pub log_color: bool,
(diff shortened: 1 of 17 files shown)


crates/rbuilder/src/live_builder/base_config.rs
```diff
@@ -398,6 +400,7 @@ impl Default for BaseConfig {
             live_builders: vec!["mgp-ordering".to_string(), "mp-ordering".to_string()],
             simulation_threads: 1,
             sbundle_mergeabe_signers: None,
+            live_builder_server_port: 8080,
```
metachris: perhaps instead of `server_port`, `http_api_port` would be more descriptive 🤔 

i wonder if `host` should also be configurable instead of fixed `127.0.0.1`

dvush:
We already have server that you can hook on that serves telemetry + other debug endpoints we can add health endpoint there too. What is the justification for not using it and for adding additional server instead?

crates/rbuilder/Cargo.toml
```diff
@@ -12,6 +12,7 @@ serde = "1.0.188"
 serde_json = "1.0.105"
 thiserror = "1.0.47"
 eyre = "0.6.8"
+actix-web = "4.9.0"
```
dvush: We already have server in the repo and its using different server.

crates/rbuilder/src/live_builder/cli.rs
```diff
@@ -80,6 +80,16 @@ pub async fn run<ConfigType: LiveBuilderConfig>(
...
+    thread::spawn(move || {
+        let sys = actix_web::rt::System::new();
+        sys.block_on(async {
+            server::start(live_builder_server_port)
+                .await
+                .map_err(|e| eprintln!("Error starting live_builder server: {:?}", e))
+                .unwrap()
+        });
+    });
     spawn_telemetry_server(
```
dvush: this is the web server I mentioned in the comment

dvush (changes requested):

crates/rbuilder/src/live_builder/base_config.rs
ZanCorDX: Last time we talked about this we decided to use a single port for everything for config simplicity.
We use LiveBuilder::with_extra_rpc for that.

ZanCorDX:
The names opaque_server/debug_server tells me nothing. I liked more "telemetry".
Maybe full_telemetry and health_check (or redacted_telemetry)?

ZanCorDX (approved):

dvush (approved):

PR #168, opened 2024-09-04
store_error feature optional.

ZanCorDX:
## 📝 Summary

This PR allows to optionally disable the store_error feature since it allows storing arbitrary data for further analysis.
Sometimes it's even used to store fully sealed blocks.

## 💡 Motivation and Context

This change is needed to avoid data leaks in a TEE context.

* [X ] Run `make lint`
* [X] Run `make test`

### DIFF
--- a/crates/rbuilder/src/utils/error_storage.rs
+++ b/crates/rbuilder/src/utils/error_storage.rs
@@ -5,7 +5,11 @@
 use crossbeam_queue::ArrayQueue;
 use lazy_static::lazy_static;
 use sqlx::{sqlite::SqliteConnectOptions, ConnectOptions, Executor, SqliteConnection};
-use std::{path::Path, time::Duration};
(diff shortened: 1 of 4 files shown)


ferranbt (approved):

crates/rbuilder/src/utils/error_storage.rs
```diff
@@ -34,18 +46,23 @@ pub async fn spawn_error_storage_writer(
     global_cancel: CancellationToken,
 ) -> eyre::Result<()> {
     let mut storage = ErrorEventStorage::new_from_path(db_path).await?;
```
liamaharon, line 48: Should this return an Err at the beginning if there is no event queue?

ZanCorDX: This is the func creating the queue!

PR #170, opened 2024-09-05
Improve metrics developer experience

liamaharon:
Currently, to add a new metric code needs to be modified in two places:
1. The big lazy_static declaring metrics
2. Inside `fn register_custom_metrics` to register it

This PR adds a macro `register_metrics` we can use in place of `lazy_static` when declaring metrics, which handles registering the declared metrics and therefore removing the requirement for `fn register_custom_metrics` entirely. 

Metrics are eagerly initialised with `ctor` so they will show up in the Prometheus metrics endpoint immediately.

I decided on this approach rather than switch to an entirely different framework like what reth uses so it would slot into the existing pattern with minimal changes.

# Usage

```rust
register_metrics! {
    pub static CURRENT_BLOCK: IntGauge = IntGauge::new("current_block", "Current Block").unwrap();

    pub static BLOCK_FILL_TIME: HistogramVec = HistogramVec::new(
        HistogramOpts::new("block_fill_time", "Block Fill Times (ms)")
            .buckets(exponential_buckets_range(1.0, 3000.0, 100)),
        &["builder_name"]
    )
    .unwrap();
}
```

expands to

```rust
lazy_static! {
    pub static ref CURRENT_BLOCK: IntGauge = {
        let metric = IntGauge::new("current_block", "Current Block").unwrap();
        REGISTRY.register(Box::new(metric.clone())).unwrap();
        metric
    };

    pub static ref BLOCK_FILL_TIME: HistogramVec = {
        let metric = HistogramVec::new(
            HistogramOpts::new("block_fill_time", "Block Fill Times (ms)")
                .buckets(exponential_buckets_range(1.0, 3000.0, 100)),
            &["builder_name"]
        ).unwrap();
        REGISTRY.register(Box::new(metric.clone())).unwrap();
        metric
    };
}

#[ctor]
fn initialize_metrics() {
    let _ = *CURRENT_BLOCK;
    let _ = *BLOCK_FILL_TIME;
}

### DIFF
--- a/crates/rbuilder/src/telemetry/metrics_macros/src/lib.rs
+++ b/crates/rbuilder/src/telemetry/metrics_macros/src/lib.rs
@@ -1,0 +1,107 @@
+extern crate proc_macro;
+use proc_macro::TokenStream;
+use quote::quote;
+use syn::{parse_macro_input, Item};
+
+/// Generates and registers the given static Prometheus metrics to a static
+/// Registry named REGISTRY.
+///
+/// This avoids the need for the caller to manually initialize the registry
+/// and call REGISTERY.register on every metric.
+///
+/// Metrics are eagerly initialized when the program starts with `ctor`,
+/// so they will show up in the Prometheus metrics endpoint immediately.
+///
+/// Note: `lazy_static::lazy_static` and `ctor::ctor` must be in scope.
+///
+/// # Example
+///
+/// ```ignore
+/// register_metrics! {
+///     pub static CURRENT_BLOCK: IntGauge = IntGauge::new("current_block", "Current Block").unwrap();
+///
+///     pub static BLOCK_FILL_TIME: HistogramVec = HistogramVec::new(
+///         HistogramOpts::new("block_fill_time", "Block Fill Times (ms)")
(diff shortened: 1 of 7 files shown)

### CODE REVIEW

**Review by ZanCorDX — APPROVED** [review]

File: crates/rbuilder/src/telemetry/metrics_macros/src/lib.rs [hunk 1]
```diff
@@ -0,0 +1,107 @@
...
+            let vis = &static_item.vis;
+            let ident = &static_item.ident;
+            let ty = &static_item.ty;
+            let expr = &static_item.expr;
+
+            // Create the static metric call REGISTER.register with it.
+            metrics.extend(quote! {
+                #vis static ref #ident: #ty = {
+                    let metric = #expr;
+                    REGISTRY.register(Box::new(metric.clone())).unwrap();
```
bertmiller: NIT: but does it make more sense to use .expect() here? That way we could get a more specific error message rather than a generic panic if something goes wrong.

bertmiller (commented):
One small comment!

ferranbt (approved):

PR #172, opened 2024-09-06
removes nonce struct in favor of bundledtxinfo

montekki:
## 📝 Summary

I've been reading through the `primitives` module and found this `Nonce` struct to be redundant and marked for deletion. This change so far does just a primitive replacement with `BundledTxInfo`. It probably also needs:

* renaming of variables from `nonce` to something more relevant
* getters on the struct to make stuff like `nonce.nonce.nonce` more readable.

Let me know what you think.

---

* [x] Run `make lint`
* [x] Run `make test`


ZanCorDX:
Can you change the names of the variables as you mentioned so instead of having nonce.nonce we have something like tx_info.nonce?
Maybe in a future PR we can rename BundledTxInfo::nonce to something like BundledTxInfo::account_nonce (open to ideas) so we don't have nonce.nonce.

PR #174, opened 2024-09-10
Sparse trie

dvush:
## Faster root hash

This pr introduces option to use sparse merkle trie implementation for the root hash calculation.

This implementation fetches relevant pieces of the trie into memory making subsequent root hash calculations faster.

On my PC reference implementation takes about 150ms to calculate hash and the new one when everything is prefetched around 5ms. 

Fetching proof for data for one uncached account takes about 0.25ms.

There are things to improve in implementation itself but the basic interface for the integration into the builder will be the same. 

## debug-bench-machine

This PR also redoes debug-bench-machine.  Now you need to stop reth node and provide external rpc for it to work. 

The reason is that reth has trie only for the last block. Previous implementation used last block to generate changes and it applied that changes to the trie after the last block. Trie would already have these changes but default reth root hash does not care about that discrepancy and it would evaluate anyway, this was used as a hack to benchmark trie even though it would give incorrect results. 

Sparse trie implemenation would not work like that because it would detect errors such as removing key that was already removed from the trie. 

The benefit of this is that we can actually check correctness of the root hash on historical block.

---

* [ ] Run `make lint`
* [ ] Run `make test`

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -2826,6 +2826,33 @@ dependencies = [
 ]
 
 [[package]]
+name = "eth-sparse-mpt"
+version = "0.1.0"
+source = "git+https://github.com/flashbots/eth-sparse-mpt?rev=664759b#664759b53d92e64d9a2e902100691b7ff4945ece"
+dependencies = [
+ "alloy-primitives 0.8.0",
+ "alloy-rlp",
+ "alloy-trie",
+ "hash-db",
+ "rayon",
+ "reth-db-api",
+ "reth-errors",
+ "reth-execution-errors",
+ "reth-provider",
+ "reth-trie",
+ "reth-trie-db",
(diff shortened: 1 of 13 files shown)


ZanCorDX:
It would be nice to have some struct to make the roothash to avoid passing so many parameters all over.
Ideally, we could have a trait and instantiate the proper version depending on the configuration so we only create what is really needed (SkipRootHash could be the dummy roothasher).

ZanCorDX (changes requested):
Shouldn't the code of
```
eth-sparse-mpt = { git = "https://github.com/dvush/eth-sparse-mpt", rev = "a89e02e" } 
```
be inside rbuilder?

ZanCorDX (approved):

PR #171, opened 2024-09-05
Dynamic log optional

ZanCorDX:
Goes on top of https://github.com/flashbots/rbuilder/pull/169.
## 📝 Summary

Allows to disable dynamic logs.

## 💡 Motivation and Context

In some contexts (eg: TEE) they might be a security leak.

---

## ✅ I have completed the following steps:

* [x] Run `make lint`
* [x] Run `make test`
* [ ] Added tests (if applicable)

PR #175, opened 2024-09-10
Filter external logs

ZanCorDX:
## 📝 Summary
This goes on top of https://github.com/flashbots/rbuilder/pull/171.

This PRs allows a new config:
```
redact_errors = true
```
to allow external systems errors to be redacted.

See https://github.com/flashbots/rbuilder/blob/filter-external-logs/docs/LOGS_PRIVACY.md for more details.

## 💡 Motivation and Context

Protect order senders.

* [x] Run `make lint`
* [x] Run `make test`

### DIFF
--- a/crates/rbuilder/src/live_builder/block_output/relay_submit.rs
+++ b/crates/rbuilder/src/live_builder/block_output/relay_submit.rs
@@ -5,13 +5,14 @@ use crate::{
         sign_block_for_relay, BLSBlockSigner, RelayError, SubmitBlockErr, SubmitBlockRequest,
     },
     primitives::mev_boost::{MevBoostRelay, MevBoostRelayID},
+    privacy::error_redactor::{ErrorRedactor, ValidationErrorRedactor},
     telemetry::{
         add_relay_submit_time, add_subsidy_value, inc_conn_relay_errors,
         inc_failed_block_simulations, inc_initiated_submissions, inc_other_relay_errors,
         inc_relay_accepted_submissions, inc_subsidized_blocks, inc_too_many_req_relay_errors,
         measure_block_e2e_latency,
     },
-    utils::error_storage::store_error_event,
+    utils::{error_storage::store_error_event, tracing::dynamic_event},
     validation_api_client::{ValidationAPIClient, ValidationError},
 };
 use ahash::HashMap;
@@ -147,6 +148,7 @@ async fn run_submit_to_relays_job(
     config: Arc<SubmissionConfig>,
     cancel: CancellationToken,
     bid_source: Arc<dyn BestBidSource>,
+    redactor: Arc<ErrorRedactor>,
 ) -> Option<BuiltBlockInfo> {
     let mut res = None;
     // first, sleep to slot time - slot_delta_to_start_submits
@@ -273,6 +275,7 @@ async fn run_submit_to_relays_job(
                 &config,
                 cancel.clone(),
                 "Dry run",
+                redactor.validation(),
             )
             .instrument(submission_span)
             .await;
@@ -303,6 +306,7 @@ async fn run_submit_to_relays_job(
                     &config,
                     cancel.clone(),
                     "Optimistic check",
+                    redactor.validation(),
                 )
                 .instrument(submission_span.clone())
                 .await
@@ -360,6 +364,7 @@ pub async fn run_submit_to_relays_job_and_metrics(
     config: Arc<SubmissionConfig>,
     cancel: CancellationToken,
     bid_source: Arc<dyn BestBidSource>,
+    redactor: Arc<ErrorRedactor>,
 ) {
     let best_bid = run_submit_to_relays_job(
         best_bid.clone(),
@@ -368,6 +373,7 @@ pub async fn run_submit_to_relays_job_and_metrics(
         config,
         cancel,
         bid_source,
+        redactor,
     )
     .await;
     if let Some(best_bid) = best_bid {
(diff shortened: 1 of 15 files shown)


crates/rbuilder/src/live_builder/block_output/relay_submit.rs
```diff
@@ -415,16 +402,29 @@ pub async fn run_submit_to_relays_job_and_metrics(
...
+/// Answers if the block was validated ok.
 async fn validate_block(
     slot_data: &MevBoostSlotData,
     signed_submit_request: &SubmitBlockRequest,
     block: SealedBlock,
     config: &SubmissionConfig,
     cancellation_token: CancellationToken,
-) -> Result<(), ValdationError> {
+    validation_use: &str,
+    redactor: Arc<dyn ValidationErrorRedactor>,
```
liamaharon, line 423: It makes me sad to see every fns params spammed with `redactor`, this feels like it will leak to absolutely everything. 

Will there ever be the need to redact some parts of the app and not others? I know you like to avoid them (I do too), but this may be a good use case for a lazy_static global, for the same reason logging libraries don't require passing the logging config to everywhere you want to log something.

crates/rbuilder/src/live_builder/block_output/relay_submit.rs
```diff
@@ -470,11 +499,11 @@ async fn submit_bid_to_the_relay(
             cancel.cancel();
         }
         Err(SubmitBlockErr::BidBelowFloor | SubmitBlockErr::PayloadAttributesNotKnown) => {
-            trace!(err = ?relay_result.unwrap_err(), "Block not accepted by the relay");
+            trace!(err = ?redactor.redact(relay_result.unwrap_err()), "Block not accepted by the relay");
```
liamaharon, line 502: Maybe we can create our own wrapper around `trace!`, `error!`, etc that adds the redaction wrapper for us, it seems like an easy thing to forget to do.

crates/rbuilder/src/mev_boost/mod.rs
```diff
@@ -225,6 +225,46 @@ pub enum RelayError {
...
+    } else if err.is_status() {
+        RedactedRequestErrorKind::Status
+    } else if err.is_body() {
+        RedactedRequestErrorKind::Body
+    } else if err.is_decode() {
+        RedactedRequestErrorKind::Decode
+    } else {
+        RedactedRequestErrorKind::Unknown
+    }
 }
```
liamaharon, line 268: Generally better to use impl From for type conversion when possible. It allows using `reqwest_err.into()` and `RedactedReqeustErrorKind::from(reqwest_err)`.

```suggestion
impl From<&reqwest::Error> for RedactedRequestErrorKind {
    fn from(err: &reqwest::Error) -> Self {
        if err.is_builder() {
            RedactedRequestErrorKind::Builder
        } else if err.is_request() {
            RedactedRequestErrorKind::Request
        } else if err.is_redirect() {
            RedactedRequestErrorKind::Redirect
        } else if err.is_status() {
            RedactedRequestErrorKind::Status
        } else if err.is_body() {
            RedactedRequestErrorKind::Body
        } else if err.is_decode() {
            RedactedRequestErrorKind::Decode
        } else {
            RedactedRequestErrorKind::Unknown
        }
    }
}
```

crates/rbuilder/src/live_builder/block_output/relay_submit.rs
ZanCorDX: I don't like singletons/global stuff (they hide dependencies) but you might be right that for these I could relax my OCDness.
Let's keep an eye on this and if this grows too much I'll go for the singleton.

crates/rbuilder/src/mev_boost/mod.rs
```diff
@@ -225,6 +225,46 @@ pub enum RelayError {
...
+    Body,
+    Decode,
+    Unknown,
+}
+
+impl std::fmt::Display for RedactedRequestErrorKind {
+    fn fmt(&self, f: &mut std::fmt::Formatter<'_>) -> std::fmt::Result {
+        write!(f, "{:?}", self)
+    }
+}
```
liamaharon, line 250: If this is a common annoyance we can use `strum` https://docs.rs/strum/latest/strum/derive.Display.html

crates/rbuilder/src/live_builder/block_output/relay_submit.rs
ZanCorDX: I thought about that but I don't know enough about macros and probably would take me a lot.
I would love to be able to put inside error! things like "if they try to log this type DON'T let it compile"

crates/rbuilder/src/privacy/mod.rs
```diff
@@ -0,0 +1,9 @@
+//! In some installations (eg: TDX) we may want to avoid ever exposing some information.
+//! This module contains all redacting stuff to achieve that.
+//! We try to keep this redacting always configurable so we can have the full data for development.
+pub mod error_redactor;
+pub mod relay;
+pub mod validation;
```
liamaharon, line 6: This is personal preference, but I'd prefer putting `relay` error redaction logic close to the relay error, and `validation` error redaction logic close to the validation error, instead of bundling all the redaction stuff in `privacy`. 

Otherwise it 
1. creates one huge module (in future maybe crate) that depends on every other module/crate in the codebase
2. imo makes the codebase less cohesive and more difficult to navigate & maintain. imagine adding a new variant to the mev_boost `RelayError`. I would prefer to stay inside the mev_boost package to impl how to redact it, instead of needing to jump back and forth between entirely different packages.

ZanCorDX: It feels like invading the other modules with some functionality they don't really care about.
But I don't have a strong position about this. If you answer "do it" I'll go with your version.

docs/LOGS_PRIVACY.md
```diff
@@ -0,0 +1,45 @@
+# Logs Privacy in rbuilder
```
metachris, line 1: Imo, ideally the docs have a date in the filename (and/or in the text), for additional context. i.e. `20240911_logs-privacy.md`. Not a strong opinion, and up to you.

PR #176, opened 2024-09-11
redact sensitive feature flag

liamaharon:
demo for @ZanCorDX 

Idea is to redact the error `Display` and `Debug` implementation where the error is defined, rather than creating new layers of wrapper we need to maintain and remember to use


.github/workflows/checks.yaml
```diff
@@ -92,13 +99,13 @@ jobs:
           echo "RUSTC_WRAPPER=sccache" >> $GITHUB_ENV
 
       - name: Build the rbuilder
-        run: cargo build
+        run: cargo build --features="${{ matrix.features }}"
```
ferranbt, line 102: Are we doubling or CI times with this? Do we need to CI test that the logs are correct?

liamaharon: No, they run as seperate jobs in parallel. Yes ci should check all features.

ZanCorDX (approved):

metachris (approved):

crates/rbuilder/src/live_builder/block_output/relay_submit.rs
```diff
@@ -474,7 +471,7 @@ async fn submit_bid_to_the_relay(
         }
         Err(SubmitBlockErr::SimError(err)) => {
             inc_failed_block_simulations();
-            error!(err = ?err, "Error block simulation fail, cancelling");
+            error!(err = ?err.clone(), "Error block simulation fail, cancelling");
```
ZanCorDX: Leaking err!

ZanCorDX (changes requested):
Did you forget to redact SubmitBlockErr?

liamaharon:
> Did you forget to redact SubmitBlockErr?

Yes good catch, I missed that that also needs redacting from your previous PR. Fixing.

ZanCorDX (approved):

Issue #180, opened 2024-09-15
Return none when draining new orders and returning an empty vec


bertmiller:
IMO `drain_new_orders()` should return _None_ if it has an empty vector. It's not very intuitive to me that it can return _Some_ with an empty vector, which lead to a bug when integrating with a new algorithm.

[Here's the relevant code](https://github.com/flashbots/rbuilder/blob/6f40cfe808fd6719c7d3da9f808c81e32f0b51ed/crates/rbuilder/src/building/block_orders/mod.rs#L90).

ZanCorDX:
Disagree.
- I don't like using None for empty vectors since it generates non uniform code and the empty vector is as valid as any other vector.
- **The semantic of the None here is very specific**. It means that you received a remove so your groups are no longer valid (at the moment of the creation of this code we had no way to unmerge groups) and instead of trying to upgrade your groupings you should restart.

PR #186, opened 2024-09-22
 returns None on empty vec

proxima424:
## 📝 Summary

Changed `drain_new_orders` implementation.
Now it returns `None` if new_orders is an empty Vec

in reference to #180


ZanCorDX:
I believe that https://github.com/flashbots/rbuilder/issues/180 is wrong, just added a comment.

PR #187, opened 2024-09-23
Parallel sealing

ZanCorDX:
## 📝 Summary

This PR replaces the SequentialSealerBidMaker for the ParallelSealerBidMaker to allow concurrent block sealing.

## 💡 Motivation and Context

Inclution rate went down when we deployed the version using SequentialSealerBidMaker.

* [ X] Run `make lint`
* [ X] Run `make test`

### DIFF
--- a/crates/rbuilder/src/live_builder/block_output/bidding/interfaces.rs
+++ b/crates/rbuilder/src/live_builder/block_output/bidding/interfaces.rs
@@ -51,7 +51,7 @@ impl Bid {
     }
 }
 
-/// Makes the actual bid (send it to the relay)
+/// Makes the actual bid (seal + send it to the relay).
 pub trait BidMaker: std::fmt::Debug {
     fn send_bid(&self, bid: Bid);
 }
--- a/crates/rbuilder/src/live_builder/block_output/bidding/parallel_sealer_bid_maker.rs
+++ b/crates/rbuilder/src/live_builder/block_output/bidding/parallel_sealer_bid_maker.rs
@@ -1,8 +1,7 @@
-use reth_primitives::format_ether;
 use std::sync::{Arc, Mutex};
 use tokio::sync::Notify;
 use tokio_util::sync::CancellationToken;
-use tracing::{error, info};
+use tracing::error;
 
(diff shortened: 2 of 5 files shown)


crates/rbuilder/src/live_builder/block_output/bidding/interfaces.rs
```diff
@@ -51,7 +51,7 @@ impl Bid {
     }
 }
 
-/// Makes the actual bid (send it to the relay)
+/// Makes the actual bid (seal + send it to the relay).
 pub trait BidMaker: std::fmt::Debug {
```
liamaharon, line 55: to be more rusty and help with the issue you've mentioned to me about finding it difficult to differentiate between trait and struct naming in rust, you could consider this:

```suggestion
/// Makes the actual bid (seal + send it to the relay).
pub trait MakeBid: std::fmt::Debug {
```

remember rust traits are best thought of as behaviours, they're not analogous to c++ interfaces. this file should be renamed to `traits.rs`.

crates/rbuilder/src/live_builder/block_output/bidding/parallel_sealer_bid_maker.rs
```diff
@@ -0,0 +1,161 @@
...
+                format_ether(bid)
+            } else {
+                "None".to_string()
+            },
+            can_add_payout_tx = block.can_add_payout_tx(),
+            tbv = format_ether(block.true_block_value().unwrap_or_default()),
+            "DX send_bid"
+        );
+        self.pending_bid.update(Bid::new(block, payout_tx_value));
+        //        self.pending_bid.update(bid);
```
liamaharon:
```suggestion
```

ZanCorDX: Ouch that info! was supposed to be removed!! Sorry! I'll fix that.

crates/rbuilder/src/live_builder/block_output/bidding/interfaces.rs
ZanCorDX: I'll think about this.

dvush (approved):

PR #193, opened 2024-09-27
Bundle hash for redistribution

dvush:
Adds bundle hash to redistribution output

output json diff:

```
<               "Bundle": {
<                 "uuid": "$UUID",
<                 "hash": "$HASH"
<               }
---
>               "Bundle": "$UUID"
```

### DIFF
--- a/crates/rbuilder/src/backtest/redistribute/mod.rs
+++ b/crates/rbuilder/src/backtest/redistribute/mod.rs
@@ -17,14 +17,15 @@ use ahash::{HashMap, HashSet};
 use alloy_primitives::utils::format_ether;
 use alloy_primitives::{Address, B256, I256, U256};
 pub use cli::run_backtest_redistribute;
-use jsonrpsee::core::Serialize;
 use rayon::prelude::*;
 use reth_chainspec::ChainSpec;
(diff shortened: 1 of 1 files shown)


ZanCorDX (approved):

ryanschneider (approved):
lgtm!  One question though: eventually we want to add additional "metadata" like the `sender` field as well (which would be either the bundle signing key or the tx sender depending on if the order is only a single tx or not).  Do we think we'd continue to add this new metadata to `ExtendedOrderId` as well?  Perhaps we could end up with/ something like:

```rust
pub enum ExtendedOrderId {
    Tx { hash: B256, sender: Address },
    Bundle { uuid: Uuid, hash: B256, sender: Address },
    ShareBundle{ hash: B256, sender: Address },
}
```

And I think `ShareBundle` might want an `tx_hash: Option<B256>` for single-tx sbundles (since one of the desires is for end users to be able to lookup their refunds via sent tx hashes, and IIUC private txs get converted to single-tx sbundles).

Anyways not advocating for doing any of this now, just wanted to get the idea down and make sure we're not backing ourselves into a corner by putting the hash in the OrderId!

PR #165, opened 2024-09-03
reth v1.0.6

dvush:
## 📝 Summary

Updates reth to v1.0.6

it was tested in backtest on 1.5k block and outputs are identical

## 💡 Motivation and Context

---

## ✅ I have completed the following steps:

* [x] Run `make lint`
* [x] Run `make test`
* [x] Added tests (if applicable)

PR #197, opened 2024-09-30
chore: bump rust version to 1.81

barnabasbusa:
1.79 no longer compiles.


rkrasiuk (approved):
https://github.com/flashbots/rbuilder/pull/165 updated reth deps which have msrv `1.80`

dvush (approved):

Issue #199, opened 2024-09-30
Run Reth alongside RBuilder in Docker


ferranbt:
The Rbuilder requires binding to the Reth database and the Reth IPC to work. That is not trivial to do with docker.

We should have a flavour of the Dockerfile (or just change the current Dockerfile) that runs both Reth and RBuilder togheter.

ryanschneider:
Cc @parithosh and @barnabasbusa

benleb:
multi-process docker images? 😕

barnabasbusa:
We have these here: https://hub.docker.com/r/ethpandaops/reth-rbuilder/tags

PR #207, opened 2024-10-02
fix(startup): Remove static_file_provider checks create_provider_factory

ryanschneider:
## 📝 Summary

This PR removes the calls to `static_file_provider().get_highest_static_file_block(..)` since it appears we no longer actively use the static file provider in our code.

## 💡 Motivation and Context

As far as I can tell, we no longer really use the static_file_provider inside the `ProviderFactoryReopener`, but the check in `create_provider_factory` causes circular dependencies related to #199 or https://github.com/ethpandaops/ethereum-package/pull/786.

---

* [x] Run `make lint`
* [x] Run `make test`


ferranbt:
Why do you say is no longer used?

dvush (changes requested):
blocking until its clear that its needed so we don't merge accidentally

PR #212, opened 2024-10-09
dont' recover signer in a hot loop

dvush:
## 📝 Summary

This is pretty serious regression that I've missed while reviewing external PR on one of the reth upgrades. 

We need to be careful to not allow something like that slip in the future.

## 💡 Motivation and Context

calling recover_signer verifies transaction signature every time we try to executing something in the vm. 

---

* [ ] Run `make lint`
* [ ] Run `make test`


ferranbt (approved):
Do you feel we can add some unit test or protection to this?

dvush:
@ferranbt I think microbenchmark for commit_tx can catch that.

PR #218, opened 2024-10-17
Generic provider

### LINKED ISSUE #216: Access DB with a generic `Provider`
The `ProviderFactory` type rbuilder uses internally for DB ops is deliberately not exposed by the reth node to services running in-process alongside it, making it impossible to access and forward the node's `ProviderFactory` to the rbuilder when creating it in-process.

To resolve this, usage of the concrete `ProviderFactory` can be replaced with a generic `Provider` trait bounded by whatever it needs from [here](https://github.com/paradigmxyz/reth/blob/661b260f6172c047e8530e2331b2c84141e03c2b/crates/storage/provider/src/traits/full.rs#L14-L24). This should allow rbuilder core to support
the provider exposed by reth when run in-process, and also a `ProviderFactory` created when the DB connection is opened independent to the node out-of-process.

Note that as well as allowing rbuilder to run in-process, this is also a necessary step to support the new reth engine (default now in reth v1.1) which does not use the `ProviderFactory`.

liamaharon:
Closes #216 

Replaces all concrete usage of `ProviderFactory` and `ProviderFactoryReopener` with generics bounded by [reth provider traits](https://github.com/paradigmxyz/reth/blob/661b260f6172c047e8530e2331b2c84141e03c2b/crates/storage/provider/src/traits/full.rs#L14-L24). This allows running rbuilder in-process using a node provider and eases swapping to a more modern provider like `BlockchainProvider` in the future. 

Note that not all trait methods can be implemented because `ProviderFactoryReopener` has no access to the blockchain tree (pending state etc.).

## DB Consistency Checks

Consistency checks have been moved out of the rbuilder core into the trait implementation for `ProviderFactoryReopener`. Inside the trait implementation, the current head block number is not known, so the reopener cannot assume that the state of the block given - 256 blocks will be available. 

To not redundantly re-check the consistency of the same block, `ProviderFactoryReopener` tracks the last block verified to have consistent state and will only re-check it when it changes. This adds 1 additional read per provider call to get the head of the chain, however I believe the cost will be insignificant.

### DIFF
--- a/crates/rbuilder/src/building/builders/merging_builder/combinator.rs
+++ b/crates/rbuilder/src/building/builders/merging_builder/combinator.rs
@@ -1,16 +1,16 @@
 use crate::{
     building::{
-        builders::block_building_helper::{BlockBuildingHelper, BlockBuildingHelperFromDB},
+        builders::block_building_helper::{BlockBuildingHelper, BlockBuildingHelperFromProvider},
         BlockBuildingContext,
     },
     roothash::RootHashConfig,
 };
 
-use crate::utils::check_provider_factory_health;
-use reth::{providers::ProviderFactory, tasks::pool::BlockingTaskPool};
-use reth_db::database::Database;
+use reth::tasks::pool::BlockingTaskPool;
+use reth_db::Database;
 use reth_payload_builder::database::CachedReads;
-use std::time::Instant;
+use reth_provider::{DatabaseProviderFactory, StateProviderFactory};
+use std::{marker::PhantomData, time::Instant};
 use time::OffsetDateTime;
 use tokio_util::sync::CancellationToken;
 use tracing::{info_span, trace};
@@ -19,8 +19,8 @@ use super::{GroupOrdering, OrderGroup};
 
 /// CombinatorContext is used for merging the best ordering from groups into final block.
 #[derive(Debug)]
-pub struct CombinatorContext<DB> {
-    provider_factory: ProviderFactory<DB>,
+pub struct CombinatorContext<P, DB> {
+    provider: P,
     root_hash_task_pool: BlockingTaskPool,
     ctx: BlockBuildingContext,
     groups: Vec<OrderGroup>,
@@ -30,12 +30,17 @@ pub struct CombinatorContext<DB> {
     coinbase_payment: bool,
     root_hash_config: RootHashConfig,
     builder_name: String,
+    phantom: PhantomData<DB>,
 }
 
-impl<DB: Database + Clone + 'static> CombinatorContext<DB> {
+impl<P, DB> CombinatorContext<P, DB>
+where
+    DB: Database + Clone + 'static,
+    P: DatabaseProviderFactory<DB> + StateProviderFactory + Clone + 'static,
+{
     #[allow(clippy::too_many_arguments)]
     pub fn new(
-        provider_factory: ProviderFactory<DB>,
+        provider: P,
         root_hash_task_pool: BlockingTaskPool,
         ctx: BlockBuildingContext,
         groups: Vec<OrderGroup>,
@@ -47,7 +52,7 @@ impl<DB: Database + Clone + 'static> CombinatorContext<DB> {
         builder_name: String,
     ) -> Self {
         CombinatorContext {
-            provider_factory,
+            provider,
             root_hash_task_pool,
             ctx,
             groups,
@@ -57,6 +62,7 @@ impl<DB: Database + Clone + 'static> CombinatorContext<DB> {
             coinbase_payment,
             root_hash_config,
             builder_name,
+            phantom: PhantomData,
         }
     }
 
@@ -85,8 +91,6 @@ impl<DB: Database + Clone + 'static> CombinatorContext<DB> {
         let build_attempt_id: u32 = rand::random();
         let span = info_span!("build_run", build_attempt_id);
         let _guard = span.enter();
-        check_provider_factory_health(self.ctx.block(), &self.provider_factory)?;
-
         let build_start = Instant::now();
 
         let mut best_orderings: Vec<GroupOrdering> = self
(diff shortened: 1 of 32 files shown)


crates/rbuilder/src/building/builders/merging_builder/combinator.rs
```diff
@@ -30,12 +30,17 @@ pub struct CombinatorContext<DB> {
...
     root_hash_config: RootHashConfig,
     builder_name: String,
+    phantom: PhantomData<DB>,
 }
 
-impl<DB: Database + Clone + 'static> CombinatorContext<DB> {
+impl<P, DB> CombinatorContext<P, DB>
+where
+    DB: Database + Clone + 'static,
+    P: DatabaseProviderFactory<DB> + StateProviderFactory + Clone + 'static,
```
ferranbt, line 39: How does this work with Rbuilder as a module? AFAIK It only receives the StateProviderFactory, not DatabaseProviderFactory<DB>.

liamaharon: It is also `DatabaseProviderFactory` https://github.com/paradigmxyz/reth/blob/661b260f6172c047e8530e2331b2c84141e03c2b/crates/storage/provider/src/traits/full.rs#L14

crates/rbuilder/src/utils/provider_factory_reopen.rs
```diff
@@ -61,13 +71,23 @@ impl<DB: Database + Clone> ProviderFactoryReopener<DB> {
...
-        &self,
-        current_block_number: u64,
-    ) -> eyre::Result<ProviderFactory<DB>> {
+    ///
+    /// If the current block number is already known at the time of calling this method, you may pass it to
+    /// avoid an additional DB lookup for the latest block number.
+    pub fn check_consistency_and_reopen_if_needed(&self) -> eyre::Result<ProviderFactory<DB>> {
+        let best_block_number = self
+            .provider_factory_unchecked()
+            .best_block_number()
```
dvush, line 80: What does best block number means and how its different from the latest block number?

dvush: What does best block number means and how its different from the latest block number?

crates/rbuilder/src/utils/provider_factory_reopen.rs
```diff
@@ -129,3 +152,168 @@ pub fn check_provider_factory_health<DB: Database>(
 
     Ok(())
 }
+
+// Implement reth db traits on the ProviderFactoryReopener, allowing generic
+// DB access. Consistency checks are required only if the operations may be
```
dvush: > Consistency checks are required only if the operations may be performed on historical state.

What do you mean by "historical state"? I thought that consistency check is mandatory in the live builder because of the bug and live builder works on the latest state.

crates/rbuilder/src/utils/provider_factory_reopen.rs
```diff
@@ -61,13 +71,23 @@ impl<DB: Database + Clone> ProviderFactoryReopener<DB> {
...
+    pub fn check_consistency_and_reopen_if_needed(&self) -> eyre::Result<ProviderFactory<DB>> {
+        let best_block_number = self
+            .provider_factory_unchecked()
+            .best_block_number()
+            .map_err(|err| eyre::eyre!("Error getting best block number: {:?}", err))?;
         let mut provider_factory = self.provider_factory.lock().unwrap();
-        if !self.testing_mode {
-            match check_provider_factory_health(current_block_number, &provider_factory) {
+
+        // Don't need to check consistency for the block that was just checked.
```
dvush, line 84: I am pretty sure that its true but it can be possible that its not.

The only place that I see that would suffer from too much checking here is nonce cache in sim tree. But even then this check can have a small impact.

I don't think that we should change it now but I marked it to remember to revisit if we see some problems with reopening.

crates/rbuilder/src/utils/provider_factory_reopen.rs
```diff
@@ -1,12 +1,20 @@
 use crate::telemetry::{inc_provider_bad_reopen_counter, inc_provider_reopen_counter};
```
dvush, line 1: Overall the approach looks good.

dvush (commented):
LGTM overall.

I've tested it for a bit and it looks good. 

I want to be sure that best block number will work in our case so we can avoid subtle bugs.

crates/rbuilder/src/utils/provider_factory_reopen.rs
liamaharon: Good catch, just learned best block is latest *finalized* block, whereas latest/last block is the canonical tip (highest in db) which I guess we want to check. I will update this.

crates/rbuilder/src/utils/provider_factory_reopen.rs
liamaharon: I understood that the bug consistency checks hack around are due to historical state becoming unavailable, so we don't need to do the checks if we want the latest state, but if that's maybe not the case lmk and I will add consistency checks for all calls.

dvush: From that comment line it looks like you only need to do the check when you work on some past state(when you work on the block which is not the latest) while in reality you need it mainly when working on the tip of the blockchain (latest state) where this bug happens the most. 

But you are correct that the bug itself is related to accessing historical state but in the following sense: when you work on the latest block state evm needs to be able to provide block hashes for the last 256 blocks and when bug happens we get zero values for some of those.

dvush (approved):
LGTM

crates/rbuilder/src/utils/provider_factory_reopen.rs
```diff
@@ -129,3 +152,168 @@ pub fn check_provider_factory_health<DB: Database>(
...
+    }
+
+    fn finalized_block_num_hash(&self) -> ProviderResult<Option<BlockNumHash>> {
+        unimplemented!("This method is not supported by ProviderFactoryReopener. Please consider using a BlockchainProvider.");
+    }
+}
+
+impl<DB: Database + Clone> StateProviderFactory for ProviderFactoryReopener<DB> {
+    fn latest(&self) -> ProviderResult<StateProviderBox> {
+        self.provider_factory_unchecked().latest()
```
dvush: The check should be added here too.

dvush: we use history_by_block hash IIRC so its not critical

liamaharon: I added it to all methods

crates/rbuilder/src/utils/provider_factory_reopen.rs
liamaharon: I added it to all methods

PR #228, opened 2024-10-25
In-process rbuilder

liamaharon:
- Refactors `LiveBuilderConfig::new_builder` to take a generic provider
- Refactors `WalletBalanceWatcher` to take a generic provider
- Adds new crate `reth-rbuilder` that runs rbuilder in-process with reth
- Changes config default ports that overlap with reth


crates/reth-rbuilder/src/main.rs
```diff
@@ -0,0 +1,129 @@
...
+    use reth::cli::Cli;
+    use reth_node_builder::EngineNodeLauncher;
+    use reth_node_ethereum::{node::EthereumAddOns, EthereumNode};
+    use reth_provider::providers::BlockchainProvider2;
+
+    reth_cli_util::sigsegv_handler::install();
+
+    if let Err(err) = Cli::<ExtraArgs>::parse().run(|builder, extra_args| async move {
+        let enable_engine2 = extra_args.experimental;
+        match enable_engine2 {
```
ferranbt, line 47: Is there any reason to use the legacy engine when running as a Reth module (out of being an experimental feature)?

liamaharon: Good q, I'm not familiar with it enough to know. Figured I'd leave it as an option since that is the default in reth 1.0.6.

ZanCorDX (approved):

PR #233, opened 2024-10-31
Allow negative coinbase profit in `Order::Tx` and `Order::Bundle`

liamaharon:
When disallowed for all `Order` varients, the coinbase account has no way to spend funds when rbuilder is building all blocks for a chain.

Instead of using saturating_sub, I also tried changing `coinbase_profit` from a `U256` to an `I256`, but ran into issues where a `U256` is required to calculate stuff like `mev_gas_price` in later logic. I'd prefer that approach if anyone knows if it is feasible.

### DIFF
--- a/crates/rbuilder/src/building/order_commit.rs
+++ b/crates/rbuilder/src/building/order_commit.rs
@@ -1011,16 +1011,11 @@ impl<'a, 'b, Tracer: SimulationTracer> PartialBlockFork<'a, 'b, Tracer> {
                 )?;
                 match res {
                     Ok(ok) => {
+                        // Builder does not sign txs in this code path, so allow negative coinbase
+                        // profit.
                         let coinbase_balance_after = self.state.balance(ctx.block_env.coinbase)?;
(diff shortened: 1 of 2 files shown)


crates/rbuilder/src/building/order_commit.rs
```diff
@@ -1012,15 +1012,8 @@ impl<'a, 'b, Tracer: SimulationTracer> PartialBlockFork<'a, 'b, Tracer> {
...
-                        let coinbase_profit = match coinbase_profit(
-                            coinbase_balance_before,
-                            coinbase_balance_after,
-                        ) {
-                            Ok(profit) => profit,
-                            Err(err) => {
-                                return Ok(Err(err));
-                            }
-                        };
+                        let coinbase_profit =
```
dvush, line 1017: Can you please add comment explaining why we don't check it for txs and bundle? Its basically because builder does not sign txs in this code path.

dvush (approved):
Thanks! I think this is a good approach.

ZanCorDX (approved):

PR #243, opened 2024-11-11
Allow zero-value payloads.

### LINKED ISSUE #242: 0-value payloads not propagated
Payloads with transactions, but whose value to the fee recipient are 0, are not propagated through the submission pipeline.  These payloads still have a nominal value to the chain, so should be considered valid and delivered to relays.

mcdee:
## 📝 Summary

Allow payloads with transactions to propagate to relays, even if the payload value is 0.

Fixes #242 

## 💡 Motivation and Context

If a payload contains transactions then it should be propagated, even if its value is 0.  Although this may not seem like a particularly useful result, it is still a valid one and sending it to the relay provides useful information and allows for special-case processing of the payload.

* [ ] Run `make lint`
  * seeing error `error: failed to run custom build command for reth-mdbx-sys v1.0.6 (https://github.com/paradigmxyz/reth?tag=v1.0.6#c228fe15)`
* [ ] Run `make test`
  * seeing error `error: failed to run custom build command for reth-mdbx-sys v1.0.6 (https://github.com/paradigmxyz/reth?tag=v1.0.6#c228fe15)`
  * There do not appear to be tests for this part of the codebase

### DIFF
--- a/crates/rbuilder/src/live_builder/block_output/relay_submit.rs
+++ b/crates/rbuilder/src/live_builder/block_output/relay_submit.rs
@@ -49,7 +49,7 @@ impl BestBlockCell {
             .as_ref()
             .map(|b| b.trace.bid_value)
             .unwrap_or_default();
-        if block.trace.bid_value > old_value {
+        if best_block.is_none() || block.trace.bid_value > old_value {
             *best_block = Some(block);
             self.block_notify.notify_one();
         }
@@ -156,6 +156,7 @@ async fn run_submit_to_relays_job(
     };
 
     let mut last_bid_value = U256::from(0);
+    let mut submitted_any_block = false;
     'submit: loop {
         if cancel.is_cancelled() {
             break 'submit res;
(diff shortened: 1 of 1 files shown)


ferranbt:
Hi @mcdee. Can you give more color on this feature? When is it required? Is it only for testing purposes? That was my impression.

mcdee:
@ferranbt pretty much as the description says: even if a payload has a nominal value of 0, if it contains transactions and is the highest value payload that can be built from the available transactions then it should be considered valid.

As to why 0 value transactions should be considered at all: not paying the (arbitrary) proposer any fees is a requirement for some transaction initiators, where they cannot make payments to the proposer without knowing who the are.  The only real solution to this is to have a 0 tip for the transaction.  And ultimately it is possible for a block to only consist of such transactions.

ZanCorDX:
Just talked with our relay expert and she said: "Relay will reject 0 eth / empty blocks at the moment" so I guess we should leave it as it is.

mcdee:
> Just talked with our relay expert and she said: "Relay will reject 0 eth / empty blocks at the moment" so I guess we should leave it as it is.

Some relays will, not all of them.

ZanCorDX:
Closing it until we are sure we can get the relays to accept this.

PR #249, opened 2024-11-15
refactor

malik672:
## 📝 Summary
main improvements include better pattern matching, more efficient sorting, and cleaner control flow.

## 💡 Motivation and Context
increased perf and refactor
---


ZanCorDX:
Sorry for the delay.
I don't think the pattern matching changes improve the readability of the code but we can take the other 2 changes if you change the PR:
```conflict_sets.sort_by_key(|set| std::cmp::Reverse(set.len()));```
```let mut profits_alone = HashMap::with_capacity(orders.len());```

malik672:
> Sorry for the delay. I don't think the pattern matching changes improve the readability of the code but we can take the other 2 changes if you change the PR: `conflict_sets.sort_by_key(|set| std::cmp::Reverse(set.len()));` `let mut profits_alone = HashMap::with_capacity(orders.len());`

The pattern matching was never because of readability, the reason is simply becasue the compiler deals with it better tbh

SozinM:
Checked out of curiosity and they produce the same asm code
https://godbolt.org/z/zEc74PjrY
https://godbolt.org/z/f8hW1qav4

malik672:
> Checked out of curiosity and they produce the same asm code https://godbolt.org/z/zEc74PjrY https://godbolt.org/z/f8hW1qav4

yeah not suprising, it should definitely generate the same assembly

you should read this: https://stackoverflow.com/questions/71806844/is-matching-on-patterns-faster-than-if-else-in-rust

ZanCorDX:
Cleaning old PRs.

PR #262, opened 2024-12-03
add `reth-rbuilder` as an artifact target in CI

sukoneck:
## 📝 Summary
- add `reth-rbuilder` as an artifact target in the `release` workflow
- run `build` job in `release` workflow inside container 
- add @sukoneck as codeowner for `/.github/` directory

## 💡 Motivation and Context
1. we need to store the binary for `reth-rbuilder` as an artifact. 

2. while I was making these changes I ran into a dependency conflict with `GLIBC` so I included this change as well (happy to separate if needed). the downside of doing this, is that it doesn't work on macos runners, so I've disabled that for now. 

---

successful run here https://github.com/flashbots/rbuilder/actions/runs/12130028687

### DIFF
--- a/.github/workflows/release.yaml
+++ b/.github/workflows/release.yaml
@@ -56,6 +56,8 @@ jobs:
     needs: extract-version
     if: ${{ github.event.inputs.build-binary == 'true' || github.event_name == 'push'}} # when manually triggered or version tagged
     runs-on: ${{ matrix.configs.runner }}
+    container:
+      image: ubuntu:22.04
     env:
       VERSION: ${{ needs.extract-version.outputs.VERSION }}
     permissions:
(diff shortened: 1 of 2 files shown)


.github/workflows/release.yaml
```diff
@@ -75,68 +70,47 @@ jobs:
...
-          fetch-depth: 0 # needed for built.rs to get GIT_HEAD_REF
-
-      # https://github.com/dtolnay/rust-toolchain
-      - name: Setup rust toolchain
-        uses: dtolnay/rust-toolchain@stable
-        with:
-          target: ${{ matrix.configs.target }}
-
-      # https://github.com/WarpBuilds/rust-cache
-      - name: Run WarpBuilds/rust-cache
```
ferranbt, line 97: How does this change affect build times?

ferranbt (approved):

.github/workflows/release.yaml
sukoneck: good question. it's not significantly different. we're building in parallel for [5-10 minutes](https://github.com/flashbots/rbuilder/actions/runs/12130028687) per binary, which is a ~1 minute longer. we can reintroduce with a bit of work if build times become an issue. 
image

PR #263, opened 2024-12-03
Disable watchdog if watchdog_timeout_sec config is 0

ferranbt:
## 📝 Summary

This PR disables the watchdog if the value from config is zero.

---

* [x] Run `make lint`
* [x] Run `make test`

### DIFF
--- a/crates/rbuilder/src/bin/dummy-builder.rs
+++ b/crates/rbuilder/src/bin/dummy-builder.rs
@@ -92,7 +92,7 @@ async fn main() -> eyre::Result<()> {
         Arc<DatabaseEnv>,
         MevBoostSlotDataGenerator,
     > {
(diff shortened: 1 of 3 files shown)


liamaharon (approved):

ZanCorDX (changes requested):
I don't like hacky constants like 0. I would prefeer to make watchdog_timeout_sec Option in the cfg. The only caveat would be that it would not be fully bacwards compatible since before not having watchdog_timeout_sec would default to 3 min but we don' really use that.

ZanCorDX:
null object pattern would be nice although for this case, it may be too much (I'll have to copy/paste your changes into multioperator for the other use of the watchdog).

ZanCorDX (approved):

PR #273, opened 2024-12-07
OFAC non critical error fix 

ZanCorDX:
## 📝 Summary

This version considers the error from the relay "block rejected because of OFAC addresses" as non critical.

## 💡 Motivation and Context

Without this change, non OFAC compliant builders would stop building blocks after sending a non-OFAC block to a OFAC complaint relay (flashbot's!).

* [X] Run `make lint`
* [X] Run `make test`

### DIFF
--- a/crates/rbuilder/src/mev_boost/mod.rs
+++ b/crates/rbuilder/src/mev_boost/mod.rs
@@ -278,7 +278,7 @@ pub enum SubmitBlockErr {
     PayloadDelivered,
     #[error("Bid below floor")]
     BidBelowFloor,
-    #[error("Simulation Error")]
+    #[error("Simulation Error: {0}")]
     SimError(String),
     #[error("RPC conversion Error")]
     /// RPC validates the submissions (eg: limit of txs) much more that our model.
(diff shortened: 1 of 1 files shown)


crates/rbuilder/src/mev_boost/mod.rs
```diff
@@ -31,6 +31,16 @@ const GZIP_CONTENT_ENCODING: &str = "gzip";
 const BUILDER_ID_HEADER: &str = "X-Builder-Id";
 const API_TOKEN_HEADER: &str = "X-Api-Token";
 
+/// We don't have nice error codes for relay errors so we have to parse looking for substrings :(
+/// Simulation error base.
+const SIM_FAILED_SUBSTRING: &str = "simulation failed";
+/// Any error containing SIM_FAILED_SUBSTRING and any of SIM_FAILED_NON_CRITICAL_ERRORS is not critical so we should not stop block building.
+const SIM_FAILED_NON_CRITICAL_ERRORS: &[&str] = &[
+    "blacklisted address", // Generated block is good but contains a blacklisted address. This happens if we don't use an OFAC blacklist but the relay does.
```
sukoneck, line 39: Nit:

Use kinder language. 

```suggestion
    "blocklisted address", // Generated block is good but contains a blocklisted address. This happens if we don't use an OFAC blocklist but the relay does.
```

sukoneck (approved):

PR #274, opened 2024-12-08
feat: Add nix flake and building `op-rbuilder` and `reth-rbuilder` via flakes

makemake-kbo:
## 📝 Summary

Adds a Nix flake that packages `op-rbuilder` and `reth-rbuilder`. Provides nightly Rust, C/C++ toolchains, and cross-platform support for development and deployment.

## 💡 Motivation and Context

A Nix flake ensures deterministic environments for both development and deployment. This is especially valuable for TEE deployments where reproducible builds are essential. The environment "just works" for developers while guaranteeing consistent deployments.

### DIFF
--- a/flake.lock
+++ b/flake.lock
@@ -1,0 +1,82 @@
+{
+  "nodes": {
+    "flake-utils": {
+      "inputs": {
+        "systems": "systems"
+      },
+      "locked": {
+        "lastModified": 1731533236,
+        "narHash": "sha256-l0KFg5HjrsfsO/JpG+r7fRrqm12kzFHyUHqHCVpMMbI=",
+        "owner": "numtide",
+        "repo": "flake-utils",
+        "rev": "11707dc2f618dd54ca8739b309ec4fc024de578b",
+        "type": "github"
+      },
+      "original": {
+        "owner": "numtide",
+        "repo": "flake-utils",
+        "type": "github"
+      }
+    },
+    "nixpkgs": {
+      "locked": {
+        "lastModified": 1733581040,
+        "narHash": "sha256-Qn3nPMSopRQJgmvHzVqPcE3I03zJyl8cSbgnnltfFDY=",
+        "owner": "NixOS",
+        "repo": "nixpkgs",
+        "rev": "22c3f2cf41a0e70184334a958e6b124fb0ce3e01",
+        "type": "github"
+      },
+      "original": {
+        "owner": "NixOS",
+        "ref": "nixos-unstable",
+        "repo": "nixpkgs",
+        "type": "github"
+      }
+    },
+    "root": {
+      "inputs": {
+        "flake-utils": "flake-utils",
+        "nixpkgs": "nixpkgs",
+        "rust-overlay": "rust-overlay"
+      }
+    },
+    "rust-overlay": {
+      "inputs": {
+        "nixpkgs": [
+          "nixpkgs"
+        ]
+      },
+      "locked": {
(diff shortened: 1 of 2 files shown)


ferranbt:
Hey! Thank you for your contribution.

We are right now using `yocto` to build the TEE attestable images so we might not use `nix` for that for the time being. Anyway, I am happy to support it initially as a dev deterministic environment.

I will loop in @MoeMahhouk to give some notes on the reproducibility, I think you have to use certain env flags to make `Reth` reproducible too.

ferranbt (approved):

MoeMahhouk:
> ## 📝 Summary
> Adds a Nix flake that packages `op-rbuilder` and `reth-rbuilder`. Provides nightly Rust, C/C++ toolchains, and cross-platform support for development and deployment.
> 
> ## 💡 Motivation and Context
> A Nix flake ensures deterministic environments for both development and deployment. This is especially valuable for TEE deployments where reproducible builds are essential. The environment "just works" for developers while guaranteeing consistent deployments.

Based on our previous experiments in trying to build the rbuilder reproducibly, we had to set SOURCE_EPOCH_DATE to the repo's commit date we are building from (see [here](https://github.com/flashbots/rbuilder?tab=readme-ov-file#reproducible-builds))
This Yocto recipe ([link](https://github.com/flashbots/meta-evm/blob/main/recipes-nodes/rbuilder/rbuilder.inc)) also does build the binary reproducibly.

I don't see in the nix build here setting SOURCE_DATE_EPOCH anywhere. My questions here are:
- How is this handled?
- Have you tried rebuilding the binary twice with no cache and checking their hash checksums to verify?
- out of curiosity, would it also build the same sha checksum on a different machine setup? (such as different linux distribution, kernel versions and standard libraries versions)

fnerdman:
@makemake-kbo would you be interested in a flashbots internship with focus on TDX infra?

makemake-kbo:
> I don't see in the nix build here setting SOURCE_DATE_EPOCH anywhere.

in Nix builds, SOURCE_DATE_EPOCH is automatically handled by the build system to ensure reproducible builds.

the only time you might care about SOURCE_DATE_EPOCH in a nix context is if you're doing development work in a nix-shell/devShell.

> out of curiosity, would it also build the same sha checksum on a different machine setup? (such as different linux distribution, kernel versions and standard libraries versions)

nix controls all build inputs deterministically. it also sandboxes the build environment too. what can cause the hash to change is:
- impure builds (access network or system state)
- libraries not handled by nix (the crates used are not handled by nix but rather by cargo. this is fine because we are relying on `cargo.lock` to provide us with deterministic crates)
- difference in arch(x86, arm, poorly configured processor feature sets on x86 which falls under the impure category)
- os differences (linux vs macos)

if the above is not invalidated, you will always get the same output hash for a package. another nice bonus is that it can help with deterministic docker builds if configured to do so.

ferranbt:
Hey @makemake-kbo. After talking with the TEE team internally, it is not in our short-term team plans to use nix to build the attestable images since we are using `yocto`. 

Supporting multiple build systems, especially one we do not have expertise in or actively use, would increase our maintenance overhead significantly. I'll close this PR, but appreciate the well-thought-out proposal!

makemake-kbo:
All good. Thanks for the review and feedback!

ferranbt:
> All good. Thanks for the review and feedback!

Thanks to you. I will keep it internally as a reference in case we want to switch to nix or experiment with it more.

PR #276, opened 2024-12-10
Improve reliability of the database provider health check

dvush:
## 📝 Summary

1. It fixes +1 error in the `check_provider_factory_health` method. We need to check last 256 blocks for the block that we are building right now and that mean we need to include latest committed block to the checking range but we were excluding that.
2. It adds additional last check to eliminate possibility of having broken database transaction. 
The way we do this check is that we look at the database provider factory which is not the db transaction but something that creates them. So there is a possibility that we can check provider factory just before it breaks. Checking it in the block builder helper makes this check on the database transaction level so there is a guarantee that the state will not be changed during the lifetime of the block builder helper.

## 💡 Motivation and Context

Database provider factory is likely to be responsible for the incorrect block that we built.

* [x] Run `make lint`
* [x] Run `make test`

### DIFF
--- a/crates/rbuilder/src/building/builders/block_building_helper.rs
+++ b/crates/rbuilder/src/building/builders/block_building_helper.rs
@@ -22,6 +22,7 @@ use crate::{
     primitives::SimulatedOrder,
     roothash::RootHashConfig,
     telemetry,
+    utils::{check_block_hash_reader_health, HistoricalBlockError},
 };
 
 use super::Block;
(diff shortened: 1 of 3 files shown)


ZanCorDX (approved):

ferranbt (approved):

dvush:
I'll benchmark how long does it take to do this check and I'll merge if its ok.

dvush:
I takes less than 100micro seconds to do this check and its pretty consistent.

PR #284, opened 2024-12-12
Process only increasing block numbers

ZanCorDX:
## 📝 Summary

Instead of processing anything that comes from the CL we only process increasing block numbers.

## 💡 Motivation and Context

I've seen some really crazy sequences from the CL (eg: repeated block numbers, going back) and the builder is not really suited right now to build simultaneous blocks (the good one and some trash).

---

* [X] Run `make lint`
* [X] Run `make test`

### DIFF
--- a/crates/rbuilder/src/live_builder/mod.rs
+++ b/crates/rbuilder/src/live_builder/mod.rs
@@ -131,6 +131,8 @@ where
     }
 
     pub async fn run(self) -> eyre::Result<()> {
+        // We keep the last block to avoid going back in time since we are now very robust about reorgs or this kind of behavior.
+        let mut last_processed_block: Option<u64> = None;
         info!("Builder block list size: {}", self.blocklist.len(),);
         info!(
             "Builder coinbase address: {:?}",
@@ -195,8 +197,13 @@ where
                 );
                 continue;
             }
+            // Allow only increasing blocks
+            if last_processed_block.map_or(false, |last_processed_block| {
+                payload.block() <= last_processed_block
+            }) {
+                continue;
+            }
             // see if we can get parent header in a reasonable time
-
             let time_to_slot = payload.timestamp() - OffsetDateTime::now_utc();
             debug!(
                 slot = payload.slot(),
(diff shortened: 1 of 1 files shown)


ferranbt (approved):

dvush (commented):
I am not sure about this one, in particular rejecting "the same block number".

IIRC CL sends 2 events per slot so it can be the case that the second one is actually "correct".

Also, is this place in the code is actually a problem? The way it works is that it just sends all events to the builder and then builder only builds on the events that have parent hash that reth hash synced. So it kind of sorted out already, even thought not directly.

ZanCorDX:
> I am not sure about this one, in particular rejecting "the same block number".
> 
> IIRC CL sends 2 events per slot so it can be the case that the second one is actually "correct".
> 
> Also, is this place in the code is actually a problem? The way it works is that it just sends all events to the builder and then builder only builds on the events that have parent hash that reth hash synced. So it kind of sorted out already, even thought not directly.

I think we can have any parent (it's just a historic block) but the real failure occurs on the root hash which, from what I undestand, can only be computed on the head. But we don't do root hash until we bid and that's too close to the slot start.
That added to the fact that I think the simulations are not really working for multiple parallel block building might hurt the good block.
Also, I was aiming to remove noisy logs.
That said, all of this comes from looking logs (it's not that easy to reproduce) so if you feel this is too risky I can leave it out.

dvush:
I think its too risky but without good argument, just on the feel level.

Also, we do wait for the parent header to be correct before even starting the block building. Maybe we can make that check harder to pass for events that we can't even build on.

dvush:
I think this can be closed, after #459 I've been tracking payloads and I don't see the need for it.
On the other hand we still miss 0.2% of slots and its unrelated to this issue so it will be handled later. 

If this issue will arise we will fix. The proper fix would be to not filter already seen payloads from CL so builder can go back if reth state also goes back. All jobs for which we can't build will be cancelled and only correct job will run. This should be done carefully so we don't run two identical jobs.
I did not see this issue though, so it might not even be needed.

Issue #302, opened 2024-12-19
Make `el_node_ipc_path` optional


ferranbt:
The `el_node_ipc_path` parameter is used to connect the EL mempool and sync it up with the builder pool. If it is not found, the builder fails. This is just another method to get order flow inside the builder, so it should be optional to enable it.

PR #314, opened 2024-12-21
Make IPC Path optional

7suyash7:
Make `el_node_ipc_path` optional in config.

#302 

* [✅ ] Run `make lint`
* [✅ ] Run `make test`

### DIFF
--- a/crates/rbuilder/src/live_builder/order_input/mod.rs
+++ b/crates/rbuilder/src/live_builder/order_input/mod.rs
@@ -17,8 +17,12 @@ use alloy_consensus::Header;
 use jsonrpsee::RpcModule;
 use parking_lot::Mutex;
 use reth_provider::StateProviderFactory;
-use std::time::Instant;
-use std::{net::Ipv4Addr, path::PathBuf, sync::Arc, time::Duration};
+use std::{
+    net::Ipv4Addr,
+    path::{Path, PathBuf},
+    sync::Arc,
+    time::Duration,
+};
 use tokio::{sync::mpsc, task::JoinHandle};
 use tokio_util::sync::CancellationToken;
 use tracing::{debug, error, info, trace, warn};
@@ -82,7 +86,7 @@ pub struct OrderInputConfig {
     /// if true -- txs with blobs are ignored
     ignore_blobs: bool,
     /// Path to reth ipc
-    ipc_path: PathBuf,
+    ipc_path: Option<PathBuf>,
     /// Input RPC port
     server_port: u16,
     /// Input RPC ip
@@ -103,7 +107,7 @@ impl OrderInputConfig {
     pub fn new(
         ignore_cancellable_orders: bool,
         ignore_blobs: bool,
-        ipc_path: PathBuf,
+        ipc_path: Option<PathBuf>,
         server_port: u16,
         server_ip: Ipv4Addr,
         serve_max_connections: u32,
@@ -123,7 +127,11 @@ impl OrderInputConfig {
     }
 
     pub fn from_config(config: &BaseConfig) -> eyre::Result<Self> {
-        let el_node_ipc_path = expand_path(config.el_node_ipc_path.clone())?;
+        let el_node_ipc_path = config
+            .el_node_ipc_path
+            .as_ref()
+            .map(|p| expand_path(p.as_path()))
+            .transpose()?;
 
         Ok(OrderInputConfig {
             ignore_cancellable_orders: config.ignore_cancellable_orders,
@@ -139,7 +147,7 @@ impl OrderInputConfig {
 
     pub fn default_e2e() -> Self {
         Self {
-            ipc_path: PathBuf::from("/tmp/anvil.ipc"),
+            ipc_path: Some(PathBuf::from("/tmp/anvil.ipc")),
             results_channel_timeout: Duration::new(5, 0),
             ignore_cancellable_orders: false,
             ignore_blobs: false,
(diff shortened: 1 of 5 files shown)


ZanCorDX:
Cleanup can't be disabled. 
Maybe https://github.com/flashbots/rbuilder/issues/302 is not so easy. It would need to trigger the cleanup from somewhere else (main block building loop?).

ferranbt:
Hey @7suyash7. @ZanCorDX is correct. I missed the cleaning part. Opened a new issue to fix that https://github.com/flashbots/rbuilder/issues/318

ferranbt:
Hey @7suyash7. Can you rebase on top of the main branch? I merged #327 and now you should be unblocked to make IPC optional.

7suyash7:
Hi, @ferranbt I just rebased on top of main. Can you please go through the changes?

crates/rbuilder/src/live_builder/order_input/txpool_fetcher.rs
```diff
@@ -24,10 +24,23 @@ pub async fn subscribe_to_txpool_with_blobs(
     results: mpsc::Sender<ReplaceableOrderPoolCommand>,
     global_cancel: CancellationToken,
 ) -> eyre::Result<JoinHandle<()>> {
-    let ipc = IpcConnect::new(config.ipc_path);
-    let provider = ProviderBuilder::new().on_ipc(ipc).await?;
-
     let handle = tokio::spawn(async move {
+        // Early return if no IPC path
```
ferranbt: I would prefer to handle this at `start_orderpool_jobs` and either run `subscribe_to_txpool_with_blobs` if IPC is set or do not do it otherwise.

ferranbt (changes requested):
Thank you for the change :). Left a comment.

crates/rbuilder/src/live_builder/order_input/txpool_fetcher.rs
```diff
@@ -24,10 +24,19 @@ pub async fn subscribe_to_txpool_with_blobs(
     results: mpsc::Sender<ReplaceableOrderPoolCommand>,
     global_cancel: CancellationToken,
 ) -> eyre::Result<JoinHandle<()>> {
-    let ipc = IpcConnect::new(config.ipc_path);
-    let provider = ProviderBuilder::new().on_ipc(ipc).await?;
-
     let handle = tokio::spawn(async move {
+        let provider = match ProviderBuilder::new()
```
ferranbt: Is this change necessary?

crates/rbuilder/src/live_builder/order_input/mod.rs
```diff
@@ -215,12 +225,18 @@ where
...
+        info!("IPC path configured, starting txpool subscription");
+        txpool_fetcher::subscribe_to_txpool_with_blobs(
+            config.clone(),
+            order_sender.clone(),
+            global_cancel.clone(),
+        )
+        .await?
+    } else {
+        info!("No IPC path configured, skipping txpool subscription");
+        tokio::spawn(async {}) // Empty task
```
ferranbt: Instead of this, can you have an array of async handles and append this one if IPC is enabled?
```
let handles = [clean_job, rpc_server]

if ipc_is_enabled {
   let txpool_fetcher = subscribe_to_txpool...
   handles.add(txpool_fetcher)
}
```

ferranbt (changes requested):

crates/rbuilder/src/live_builder/order_input/txpool_fetcher.rs
7suyash7: No, I removed it.

crates/rbuilder/src/live_builder/order_input/mod.rs
7suyash7: Yes, can you please check the new commit

crates/rbuilder/src/live_builder/order_input/txpool_fetcher.rs
```diff
@@ -24,10 +24,20 @@ pub async fn subscribe_to_txpool_with_blobs(
     results: mpsc::Sender<ReplaceableOrderPoolCommand>,
     global_cancel: CancellationToken,
 ) -> eyre::Result<JoinHandle<()>> {
-    let ipc = IpcConnect::new(config.ipc_path);
-    let provider = ProviderBuilder::new().on_ipc(ipc).await?;
-
     let handle = tokio::spawn(async move {
+        let provider = ProviderBuilder::new()
+            .on_ipc(IpcConnect::new(config.ipc_path.unwrap()))
```
ferranbt: Is this change still necessary? We can fail with `?` and do not need to cancel `global_cancel`.

ferranbt (changes requested):
One last comment and it should be ready to merge.

crates/rbuilder/src/live_builder/order_input/txpool_fetcher.rs
7suyash7: Hi, using `?` here would mean that we would have to change the return type of the tokio::spawn to a `Result` instead of `()`, I looked at other spawned tasks (clean_job, rpc_server) and they all handle errors internally, so I thought I should do the same...
Also, for `global_cancel` - if the user specified a path to the IPC and it failed to connect, is shutting down not appropriate?
I'm happy to make both changes, just asking some questions for clarity

ferranbt: That happens because you put the ProviderBuilder inside the spawn task. In the original code the provider is created outside. Would reverting to that work?

On a separated note, try to not use `unwrap` in the config ipc path.

7suyash7: Got it, can you please check the latest commit? Rust patterns are somewhat new and confusing to me, thanks for the guidance!

ferranbt (approved):

PR #316, opened 2024-12-25
refactor(op-rbuilder): simplify OpRbuilderPayloadBuilder implementation

bendanzhentan:
## 💡 Motivation and Context

`OpRbuilderPayloadBuilder` includes two parameters `compute_pending_block` and `evm_config`. The construction pattern should either use builder-mode for both or follow new-mode for both. I make it to new-mode.

---

* [x] Run `make lint`
* [x] Run `make test`

### DIFF
--- a/crates/op-rbuilder/node/src/node.rs
+++ b/crates/op-rbuilder/node/src/node.rs
@@ -281,9 +281,10 @@ impl OpRbuilderPayloadServiceBuilder {
 
         Evm: ConfigureEvm<Header = Header>,
     {
-        let payload_builder =
-            op_rbuilder_payload_builder::OpRbuilderPayloadBuilder::new(evm_config)
-                .set_compute_pending_block(self.compute_pending_block);
+        let payload_builder = op_rbuilder_payload_builder::OpRbuilderPayloadBuilder::new(
+            evm_config,
+            self.compute_pending_block,
+        );
         let conf = ctx.payload_builder_config();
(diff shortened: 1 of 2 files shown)


liamaharon (changes requested):
Thank you for your PR!

While the reth apis have room for improvement, I would prefer to maintain consistency with their `PayloadBuilder`s, which use the builder pattern.

Introducing a deviation here is likely to make upgrading reth versions more time consuming, and our api potentially confusing to developers who are already familiar with reth. 

I suggest you open a similar pr to reth updating the api on their `PayloadBuilder`s, and we will follow suit :)

bendanzhentan:
> Thank you for your PR!
> 
> While the reth apis have room for improvement, I would prefer to maintain consistency with their `PayloadBuilder`s, which use the builder pattern.
> 
> Introducing a deviation here is likely to make upgrading reth versions more time consuming, and our api potentially confusing to developers who are already familiar with reth.
> 
> I suggest you open a similar pr to reth updating the api on their `PayloadBuilder`s, and we will follow suit :)

Thanks for your kindly suggestion @liamaharon , and sorry for my late reply. Let's close it at this moment.

ferranbt:
Hey @bendanzhentan. Feel free to join our [Telegram channel](https://t.me/flashbots_rbuilder) if you want to have some convos about any other changes that you might be interested in doing.

PR #320, opened 2024-12-31
Fail if no relays are configured

### LINKED ISSUE #304: If `dry_run` is not enabled and there are no relays set, the builder should fail
If the builder is not in `dry_run` mode and the relays list is empty, the builder starts and it does nothing. It needs the relay to generate the mev slot data in `MevSlotDataGenerator`.

Currently, we rely on watchdog to stop the process eventually if there are no slots found, but we are trying to remove this behaviour.

### LINKED ISSUE #319: The L1 rbuilder should not run if no relays are enabled and not in dry run mode
If not in `dry_run` mode and there are no relays configured, the builder should fail.

ferranbt:
## 📝 Summary

Closes #304 

## 💡 Motivation and Context

---

* [ ] Run `make lint`
* [ ] Run `make test`

### DIFF
--- a/crates/rbuilder/src/live_builder/config.rs
+++ b/crates/rbuilder/src/live_builder/config.rs
@@ -278,6 +278,10 @@ impl L1Config {
         );
 
         let relays = self.create_relays()?;
+        if !self.dry_run && relays.is_empty() {
+            eyre::bail!("No relays provided");
+        }
+
         let sink_factory: Box<dyn BuilderSinkFactory> = Box::new(RelaySubmitSinkFactory::new(
             submission_config,
             relays.clone(),
--- a/crates/rbuilder/src/live_builder/config.rs
+++ b/crates/rbuilder/src/live_builder/config.rs
@@ -278,7 +278,7 @@ impl L1Config {
         );
 
         let relays = self.create_relays()?;
-        if !self.dry_run && relays.is_empty() {
+        if relays.is_empty() {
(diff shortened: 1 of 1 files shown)


ZanCorDX:
Why do you check for ```!self.dry_run && relays.is_empty()```?
Wouldn't it also fail in dryrun mode with no relays since it can't get the slot info on tha main loop?

ferranbt:
True, I have to do only `relays.is_empty()` only.

crates/rbuilder/src/live_builder/config.rs
```diff
@@ -278,6 +278,10 @@ impl L1Config {
         );
 
         let relays = self.create_relays()?;
+        if relays.is_empty() {
+            eyre::bail!("No relays provided");
+        }
```
liamaharon, line 283: `op-rbuilder` doesn't require relays. With all the assumptions the `new_builder` method makes about L1 specific config, maybe it should be renamed `new_l1_builder`?

liamaharon (approved):

crates/rbuilder/src/live_builder/config.rs
ferranbt: Agree, my goal is to end up with a chain-agnostic core block building and then make LiveBuilder the specific instance that "wires" that block builder for the Ethereum L1 chain.

ZanCorDX (approved):

PR #321, opened 2024-12-31
Add list of default relays and config to enable them

### LINKED ISSUE #305: Hardcode relay addresses in the builder
Right now, the user has to supply a list of configuration for all the builders he wants to connect to:

```config.toml
[[relays]]
name = "flashbots"
url = "https://0xac6e77dfe25ecd6110b8e780608cce0dab71fdd5ebea22a16c0205200f2f8e2e3ad3b71d3499c54ad14d6c21b41a37ae@boost-relay.flashbots.net"
priority = 0
use_ssz_for_submit = false
use_gzip_for_submit = false
```

Instead, we should built-in these configurations inside the rbuilder and let the user choose which one wants to use.

ferranbt:
## 📝 Summary

Closes #305 

## 💡 Motivation and Context

---

* [ ] Run `make lint`
* [ ] Run `make test`


ferranbt:
There is still something missing here because I cannot merge with the default config.

crates/rbuilder/src/live_builder/config.rs
```diff
@@ -169,9 +173,47 @@ impl L1Config {
     }
 
     pub fn create_relays(&self) -> eyre::Result<Vec<MevBoostRelay>> {
+        let mut relays = DEFAULT_RELAYS.to_vec();
```
ZanCorDX: Wouldn't it feel more natural/easy to use a map string->cfg instead of vector + find name?

crates/rbuilder/src/primitives/mev_boost.rs
```diff
@@ -11,7 +12,8 @@ pub type MevBoostRelayID = String;
 #[serde(deny_unknown_fields)]
 pub struct RelayConfig {
     pub name: String,
-    pub url: String,
+    pub url: Option<String>,
```
ZanCorDX: A RelayConfig with no url makes no sense. Why the Option?

ferranbt: The idea is to be able to update one of the default relays like:

```
relays = ["flashbots"]

[[relays]]
name = "flashbots"
use_gzip_for_submit = true
```
But this cannot be done because `url` is required.

ZanCorDX (approved):

PR #323, opened 2024-12-31
Add option pattern for Optimistic Relay mode

ferranbt:
## 📝 Summary

This PR introduces a new `OptimisitcConfig` for the relayer with an option pattern for its fields. This replaces the current `optimistic_enabled` boolean to signal whether to use the subset of Optimistic configs or not.

## 💡 Motivation and Context

---

* [ ] Run `make lint`
* [ ] Run `make test`

### DIFF
--- a/crates/rbuilder/src/live_builder/block_output/relay_submit.rs
+++ b/crates/rbuilder/src/live_builder/block_output/relay_submit.rs
@@ -106,6 +106,12 @@ pub struct SubmissionConfig {
     pub bid_observer: Box<dyn BidObserver + Send + Sync>,
 }
 
+/// Configuration for optimistic block submission to relays.
+///
+/// For optimistic relays when bid_value < max_bid_value:
+/// - If prevalidate_optimistic_blocks=true: Validate first, then submit with optimistic key
+/// - If prevalidate_optimistic_blocks=false: Submit directly with optimistic key
+/// Otherwise uses normal submission path.
 #[derive(Debug, Clone)]
 pub struct OptimisticConfig {
     pub signer: BLSBlockSigner,
--- a/crates/rbuilder/src/live_builder/block_output/relay_submit.rs
+++ b/crates/rbuilder/src/live_builder/block_output/relay_submit.rs
@@ -102,12 +102,15 @@ pub struct SubmissionConfig {
     pub dry_run: bool,
     pub validation_api: ValidationAPIClient,
 
-    pub optimistic_enabled: bool,
-    pub optimistic_signer: BLSBlockSigner,
-    pub optimistic_max_bid_value: U256,
-    pub optimistic_prevalidate_optimistic_blocks: bool,
-
+    pub optimisitic_config: Option<OptimisticConfig>,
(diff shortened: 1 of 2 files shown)


crates/rbuilder/src/live_builder/block_output/relay_submit.rs
```diff
@@ -102,14 +102,17 @@ pub struct SubmissionConfig {
...
+    pub optimisitic_config: Option<OptimisticConfig>,
     pub bid_observer: Box<dyn BidObserver + Send + Sync>,
 }
 
+#[derive(Debug, Clone)]
+pub struct OptimisticConfig {
+    pub signer: BLSBlockSigner,
+    pub max_bid_value: U256,
+    pub prevalidate_optimistic_blocks: bool,
+}
```
liamaharon: Some doc comments on what the optimistic config and its behavior would be nice.

crates/rbuilder/src/live_builder/block_output/relay_submit.rs
```diff
@@ -102,14 +102,17 @@ pub struct SubmissionConfig {
     pub dry_run: bool,
     pub validation_api: ValidationAPIClient,
 
-    pub optimistic_enabled: bool,
-    pub optimistic_signer: BLSBlockSigner,
-    pub optimistic_max_bid_value: U256,
-    pub optimistic_prevalidate_optimistic_blocks: bool,
-
+    pub optimisitic_config: Option<OptimisticConfig>,
```
liamaharon: Previously `optimistic` config fields were mandatory, what's the rationale making them optional?

crates/rbuilder/src/live_builder/block_output/relay_submit.rs
```diff
@@ -274,11 +296,13 @@ async fn run_submit_to_relays_job(
             );
         }
 
-        if submission_optimistic {
-            let can_submit = if config.optimistic_prevalidate_optimistic_blocks {
+        if let Some(optimistic_signed_submission) = &optimistic_signed_submission {
+            let optimistic_config = optimistic_config.expect("optimistic config is not None");
```
liamaharon: `expect` always makes me a bit nervous. it might hold true for now, but it's not guaranteed to hold in the future.

maybe ln 257 can return `Some((res, optimistic_config))`, which will bring the safely unwrapped config into scope here?

crates/rbuilder/src/live_builder/block_output/relay_submit.rs
ferranbt: The relay optimistic functionality is optional, only enabled if `optimistic_enabled` is set to true. When it is set to false, those fields have default null values. That to me is kind of the same as an Option config.

crates/rbuilder/src/live_builder/block_output/relay_submit.rs
ferranbt: That is a nice pattern :) Fixed!

ZanCorDX (approved):

PR #325, opened 2024-12-31
Add blocklist struct

ferranbt:
## 📝 Summary

This PR replaces raw `HashSet<Address>` usages throughout the codebase with a dedicated BlockList struct that encapsulates the blocklist functionality in a single location. The struct implements a custom deserialize operation that maintains backward compatibility: if a blocklist file exists it parses the addresses, otherwise it falls back to an empty blocklist. This behavior is fully unit tested.

---

* [x] Run `make lint`
* [x] Run `make test`


crates/rbuilder/src/blocklist/mod.rs
```diff
@@ -0,0 +1,148 @@
...
+        })
+    }
+
+    pub fn len(&self) -> usize {
+        self.list.len()
+    }
+
+    pub fn contains(&self, address: &Address) -> bool {
+        self.list.contains(address)
+    }
```
liamaharon: Might want to impl

```rust
impl Deref for BlockList {
    type Target = HashSet<Address>;

    fn deref(&self) -> &Self::Target {
        &self.list
    }
}
```

instead of the inner methods individually

crates/rbuilder/src/blocklist/mod.rs
```diff
@@ -0,0 +1,148 @@
...
+    #[cfg(test)]
+    fn add(&mut self, address: Address) {
+        self.list.insert(address);
+    }
+}
+
+impl From<PathBuf> for BlockList {
+    fn from(path: PathBuf) -> Self {
+        Self::from_file(path).unwrap_or_else(|_| Self::new())
+    }
```
liamaharon: I would expect this type of method to return an error if the path doesn't exist instead of an empty list. 
```suggestion
    fn try_from(path: PathBuf) -> Result<Self, ...> {
        Self::from_file(path).unwrap_or_else(|_| Self::new())
    }
```

liamaharon (commented):
Not super clear to me if the abstraction is needed, but if there is agreement it is then looks generally good!

crates/rbuilder/src/blocklist/mod.rs
ferranbt: Added Deref but not sure what changes I have to make to the len and contains functions.

ferranbt:
> Not super clear to me if the abstraction is needed, but if there is agreement it is then looks generally good!

There are three benefits for this abstraction:
- It eliminates the need to repeatedly instantiate HashSet<Address> throughout the codebase when working with blocklists
- It provides a dedicated location for potential future enhancements, such as implementing hot reloading of the blocklist
- It handles the blocklist deserialization at config load time rather than requiring a runtime blocklist() function.

crates/rbuilder/src/blocklist/mod.rs
liamaharon: You can remove them now :)

ZanCorDX:
Not sure if it's needed right now, but it's done and might be useful in the future.
I think DerefMut may be too much. I can give you Deref as syntax sugar for accessing the HashSet, but DerefMut allows changing it...

ZanCorDX (approved):
Do we really need DerefMut?

liamaharon (approved):

config-live-example.toml
```diff
@@ -20,8 +20,6 @@ jsonrpc_server_ip = "0.0.0.0"
 el_node_ipc_path = "/tmp/reth.ipc"
 extra_data = "⚡🤖"
 
-blocklist_file_path = "./blocklist.json"
```
metachris: please leave this in

PR #326, opened 2024-12-31
refactor: Use Ipv4Addr directly in config instead of Option<String>

ferranbt:
## 📝 Summary

Replace Optional String IP configuration fields with direct Ipv4Addr values since the Option type wasn't meaningful - these fields always resolved to a default IP (0.0.0.0) when None or when parsing failed. This change makes the config more type-safe and better reflects that IP addresses are required fields with sensible defaults.

---

* [ ] Run `make lint`
* [ ] Run `make test`


liamaharon (approved):
Looks good!

PR #327, opened 2024-12-31
Notify orderpool cleanup without IPC

### LINKED ISSUE #318: Make orderpool cleanup independent of IPC path
Currently, the orderpool cleanup functionality requires an IPC path to be configured, as it uses this connection to subscribe to new blocks and trigger the cleanup process. This makes the IPC path a mandatory requirement when other parts of the system already act on new blocks, and we can reuse those code paths instead of having multiple methods to track new blocks.

### Proposed Solution
Implement a subscription system that allows components to subscribe to new block updates. When a LiveBuilder receives and validates a new block (including state validation), it can emit an event that triggers all subscribed components. This way, the orderpool cleanup can subscribe to these events instead of requiring its own IPC connection.

ferranbt:
## 📝 Summary

Closes #318. This PR uses the block building flow to notify the order pool to clean old transactions instead of using the IPC path of the EL node.

## 💡 Motivation and Context

---

* [ ] Run `make lint`
* [ ] Run `make test`


ZanCorDX (approved):

crates/rbuilder/src/live_builder/order_input/mod.rs
```diff
@@ -295,3 +298,54 @@ pub fn expand_path(path: PathBuf) -> eyre::Result<PathBuf> {
...
+    let handle = tokio::spawn(async move {
+        info!("Clean orderpool job: started");
+
+        while let Some(header) = header_receiver.recv().await {
+            let block_number = header.number;
+            set_current_block(block_number);
+            let state = match provider_factory.latest() {
+                Ok(state) => state,
+                Err(err) => {
+                    error!("Failed to get latest state: {}", err);
```
dvush, line 324: please, do logs like this

`error!(?err, "Message")`

This is how its done in the rest of the codebase and it makes the errors appear as a field in the json logs instead of the message.

PR #328, opened 2024-12-31
Faheelsattar/add gen bls flag

### LINKED ISSUE #307: Add a sub-command to generate a BLS address
Add a new `rbuilder gen-bls` sub-command to generate a BLS address. As already described in #306, a common pain point is that the user needs a valid BLS address for the coinbase which might not be handy to generate.

faheelsattar:
## 📝 Summary

Adds `gen-bls` flag to rbuilder for generating BLS address
Closes #307

* [x] Run `make lint`
* [x] Run `make test`


crates/rbuilder/src/utils/bls.rs
```diff
@@ -0,0 +1,22 @@
+use blst::min_pk::SecretKey;
+use rand::{self, RngCore};
+
+pub fn generate_random_bls_address() -> [u8; 48] {
+    let mut rng = rand::thread_rng();
+    let mut ikm = [0u8; 32];
+    rng.fill_bytes(&mut ikm);
+    let sk = SecretKey::key_gen(&ikm, &[]).unwrap();
```
ferranbt: There is a `SecretKey::random` function already. Could you use that?

faheelsattar: Is there? cuz i couldn't find any. They have random address generation example in their [README](https://github.com/supranational/blst/blob/master/bindings/rust/README.md)

ferranbt: There is [this](https://github.com/ralexstokes/ethereum-consensus/blob/8fbd8a53dca0170bedeca40a92ee70fd48c4615b/ethereum-consensus/src/crypto/bls.rs#L443) one.

faheelsattar: ahh gotchu

ferranbt (approved):

PR #335, opened 2025-01-02
Faheelsattar/random coinbase address

### LINKED ISSUE #306: Generate a random coinbase address if none found
A paint point for the `rbuilder` is that it requires a coinbase address to work. If no `coinbase` address is set, we should show a warning on the logs and default to a random one. The only caveat is that if the user were to win any blocks, he could not have access to the winnings. I think this is a fair tradeoff.

faheelsattar:
## 📝 Summary

Generates a random coinbase address if none was set
Closes #306

## 💡 Motivation and Context
---

* [x] Run `make lint`
* [x] Run `make test`

### DIFF
--- a/crates/rbuilder/src/live_builder/base_config.rs
+++ b/crates/rbuilder/src/live_builder/base_config.rs
@@ -285,9 +285,7 @@ impl BaseConfig {
 
     pub fn coinbase_signer(&self) -> eyre::Result<Signer> {
         if let Some(secret_key) = &self.coinbase_secret_key {
-            if let Ok(resolved_key) = secret_key.value() {
-                return coinbase_signer_from_secret_key(&resolved_key);
-            }
+            return coinbase_signer_from_secret_key(&secret_key.value()?);
         }
         warn!("No coinbase secret key provided. A random key will be generated.");
         warn!(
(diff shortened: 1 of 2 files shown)


crates/rbuilder/src/live_builder/base_config.rs
```diff
@@ -284,7 +284,17 @@ impl BaseConfig {
     }
 
     pub fn coinbase_signer(&self) -> eyre::Result<Signer> {
-        coinbase_signer_from_secret_key(&self.coinbase_secret_key.value()?)
+        if let Some(secret_key) = &self.coinbase_secret_key {
+            if let Ok(resolved_key) = secret_key.value() {
```
ferranbt:
```suggestion
           coinbase_signer_from_secret_key(secret_key.value()?)
```
We should fail if parsing the user-supplied secret key fails.

ferranbt (changes requested):
A few nitpicks.

liamaharon (commented):
This feels like a potential footgun.

Is there a legitimate reason an L1 builder would not want to specify the coinbase address?

ferranbt (approved):

ferranbt (approved):

PR #337, opened 2025-01-04
generate random relay key when none found

### LINKED ISSUE #333: Generate a random relay key if none found
Right now, if the user does not set a relay key the builder fails. Instead, we should set a random key.

faheelsattar:
## 📝 Summary
Generates a random relay secret key when none is set
Closes #333

## 💡 Motivation and Context
---

* [x] Run `make lint`
* [x] Run `make test`

### DIFF
--- a/crates/rbuilder/src/live_builder/config.rs
+++ b/crates/rbuilder/src/live_builder/config.rs
@@ -55,6 +55,8 @@ use reth_provider::{
     BlockReader, DatabaseProviderFactory, HeaderProvider, StateProviderFactory,
     StaticFileProviderFactory,
 };
+use revm_primitives::hex;
+use secp256k1::SecretKey;
 use serde::Deserialize;
 use serde_with::{serde_as, OneOrMany};
 use std::{
@@ -64,7 +66,7 @@ use std::{
     sync::Arc,
     time::Duration,
 };
-use tracing::info;
+use tracing::{info, warn};
 use url::Url;
 
 /// We initialize the wallet with the last full day. This should be enough for any bidder.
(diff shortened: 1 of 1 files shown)


crates/rbuilder/src/live_builder/config.rs
```diff
@@ -267,6 +270,17 @@ impl L1Config {
...
+
+    fn get_relay_secret_key(&self) -> String {
+        if let Some(secret_key) = &self.relay_secret_key {
+            if let Ok(resolved_key) = secret_key.value() {
+                return resolved_key;
+            }
+        }
+        warn!("No relay secret key provided. A random key will be generated.");
+        let secret_key = SecretKey::new(&mut rand::thread_rng());
+        hex::encode(secret_key.secret_bytes())
```
ferranbt: Do not convert to string to convert back to BLSBlockSigner

crates/rbuilder/src/live_builder/config.rs
```diff
@@ -267,6 +270,17 @@ impl L1Config {
         ));
         Ok((sink_factory, relays))
     }
+
+    fn get_relay_secret_key(&self) -> String {
```
ferranbt: Please inline this in `submission_config`.

crates/rbuilder/src/live_builder/config.rs
```diff
@@ -267,6 +270,17 @@ impl L1Config {
...
+        if let Some(secret_key) = &self.relay_secret_key {
+            if let Ok(resolved_key) = secret_key.value() {
```
ferranbt: Handle the error case with ?

ferranbt (commented):
Thank you for the PR. I left a few notes.

crates/rbuilder/src/live_builder/config.rs
faheelsattar: passing `SecretKey` directly now

ferranbt (approved):

PR #338, opened 2025-01-04
fix: correctly provision # of threads

bertmiller:
## 📝 Summary

This fixes a bug where we only use 1 thread where the intention was to provision a number of threads.

## 💡 Motivation and Context

This will make resolving conflicts in the parallel builder much more efficient.
---

* [ ✅] Run `make lint`
* [ ✅] Run `make test`


liamaharon (approved):

ZanCorDX (changes requested):
It makes no sense to create a ThreadPool, just use thread::spawn.

bertmiller:
> It makes no sense to create a ThreadPool, just use thread::spawn.

Thanks, fixed

ZanCorDX (approved):

PR #347, opened 2025-01-07
refactor: `TransactionSignedEcRecoveredWithBlobs` constructor api

liamaharon:
Refactors the `TransactionSignedEcRecoveredWithBlobs` constructor api.

## Adds `new` constructor
```rust
    /// Create new with an optional blob sidecar.
    ///
    /// Warning: It is the caller's responsibility to check if a tx has blobs.
    /// This fn will return an Err if it is passed an eip4844 without blobs.
    pub fn new(
        tx: TransactionSignedEcRecovered,
        blob_sidecar: Option<BlobTransactionSidecar>,
        metadata: Option<Metadata>,
    ) -> Result<Self, TxWithBlobsCreateError>
```

## `new_no_blobs` return val 
Change return type from `Option<Self>` to `Result<Self, TxWithBlobsCreateError>` to better reflect behavior

## Adds `try_from` for `TransactionSignedEcRecovered`
```rust
    /// Try to create a [`TransactionSignedEcRecoveredWithBlobs`] from a
    /// [`TransactionSignedEcRecovered`] and reth pool.
    ///
    /// The pool is required because [`TransactionSignedEcRecovered`] on its
    /// own does not contain blob information, it is required to fetch the blob.
    ///
    /// Unfortunately we need to pass the entire pool, because the blob store
    /// is not part of the pool's public api.
    pub fn try_from_tx_without_blobs_and_pool<V, T, S>(
        tx: TransactionSignedEcRecovered,
        pool: Pool<V, T, S>,
    ) -> Result<Self, TxWithBlobsCreateError>
```

### DIFF
--- a/crates/rbuilder/src/primitives/mod.rs
+++ b/crates/rbuilder/src/primitives/mod.rs
@@ -460,7 +460,7 @@ impl TransactionSignedEcRecoveredWithBlobs {
         blob_sidecar: Option<BlobTransactionSidecar>,
         metadata: Option<Metadata>,
     ) -> Result<Self, TxWithBlobsCreateError> {
-        if tx.transaction.blob_versioned_hashes().is_some() {
+        if tx.transaction.blob_versioned_hashes().is_some() && blob_sidecar.is_none() {
             Err(TxWithBlobsCreateError::Eip4844MissingBlobSidecar)
         } else {
             Ok(Self {
--- a/crates/rbuilder/src/primitives/mod.rs
+++ b/crates/rbuilder/src/primitives/mod.rs
@@ -15,6 +15,10 @@ use alloy_eips::{
 use alloy_primitives::{keccak256, Address, Bytes, TxHash, B256, U256};
 use derivative::Derivative;
 use integer_encoding::VarInt;
+use reth::transaction_pool::{
+    BlobStore, BlobStoreError, EthPooledTransaction, Pool, TransactionOrdering, TransactionPool,
+    TransactionValidator,
+};
 use reth_primitives::{
     kzg::{BYTES_PER_BLOB, BYTES_PER_COMMITMENT, BYTES_PER_PROOF},
     PooledTransactionsElement, TransactionSigned, TransactionSignedEcRecovered,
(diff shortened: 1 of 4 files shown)


crates/rbuilder/src/primitives/mod.rs
```diff
@@ -440,18 +444,60 @@ pub enum RawTxWithBlobsConvertError {
...
+                blobs_sidecar: Arc::new(blob_sidecar.unwrap_or_default()),
+                metadata: metadata.unwrap_or_default(),
+            })
         }
-        Some(Self::new_for_testing(tx))
     }
 
-    /// Creates a Self with empty blobs sidecar. No consistency check is performed,
+    /// Shorthand for `new(tx, None, None)`
+    pub fn new_no_blobs(tx: TransactionSignedEcRecovered) -> Result<Self, TxWithBlobsCreateError> {
```
ZanCorDX, line 483: Why did you remove the check:
```
        if tx.transaction.blob_versioned_hashes().is_some() {
            return None;
        }
        
```
?

ZanCorDX (changes requested):

crates/rbuilder/src/primitives/mod.rs
```diff
@@ -440,18 +444,60 @@ pub enum RawTxWithBlobsConvertError {
...
+    /// Warning: It is the caller's responsibility to check if a tx has blobs.
+    /// This fn will return an Err if it is passed an eip4844 without blobs.
+    pub fn new(
+        tx: TransactionSignedEcRecovered,
+        blob_sidecar: Option<BlobTransactionSidecar>,
+        metadata: Option<Metadata>,
+    ) -> Result<Self, TxWithBlobsCreateError> {
         if tx.transaction.blob_versioned_hashes().is_some() {
-            return None;
+            Err(TxWithBlobsCreateError::Eip4844MissingBlobSidecar)
```
ZanCorDX: Shouldn't the check be something like tx.transaction.blob_versioned_hashes().is_some() != blob_sidecar.is_some() ?
This looks like it will fail for every tx with blobs.

ZanCorDX (changes requested):

crates/rbuilder/src/primitives/mod.rs
liamaharon: It is checked in `new`, this fn is now just for shorthand convenience.

crates/rbuilder/src/primitives/mod.rs
liamaharon: Good catch, thanks! Fixed.

ZanCorDX (approved):

PR #349, opened 2025-01-07
Stop spawn_clean_orderpool_job on shutdown

ferranbt:
## 📝 Summary

Bug introduced in #327. Rbuilder would not stop with `Ctrl-C` because it would be waiting for `spawn_clean_orderpool_job` to finish. But, `spawn_clean_orderpool_job` would not handle the cancellation token.

## 💡 Motivation and Context

---

* [x] Run `make lint`
* [x] Run `make test`

### DIFF
--- a/crates/rbuilder/src/live_builder/order_input/mod.rs
+++ b/crates/rbuilder/src/live_builder/order_input/mod.rs
@@ -334,33 +334,41 @@ where
     let handle = tokio::spawn(async move {
         info!("Clean orderpool job: started");
 
-        while let Some(header) = header_receiver.recv().await {
-            let block_number = header.number;
-            set_current_block(block_number);
-            let state = match provider_factory.latest() {
-                Ok(state) => state,
-                Err(err) => {
-                    error!("Failed to get latest state: {}", err);
-                    // @Metric error count
-                    continue;
+        loop {
+            tokio::select! {
+                Some(header) = header_receiver.recv() => {
+                    let block_number = header.number;
+                    set_current_block(block_number);
+                    let state = match provider_factory.latest() {
+                        Ok(state) => state,
+                        Err(err) => {
+                            error!("Failed to get latest state: {}", err);
+                            // @Metric error count
+                            continue;
+                        }
+                    };
+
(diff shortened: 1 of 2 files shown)


metachris (approved):

crates/rbuilder/src/live_builder/order_input/mod.rs
```diff
@@ -334,33 +334,41 @@ where
...
-                Ok(state) => state,
-                Err(err) => {
-                    error!("Failed to get latest state: {}", err);
-                    // @Metric error count
-                    continue;
-                }
-            };
+        loop {
+            tokio::select! {
+                Some(header) = header_receiver.recv() => {
```
ZanCorDX: What happens is header_receiver.recv() gives None?

ZanCorDX (changes requested):

crates/rbuilder/src/live_builder/order_input/mod.rs
ferranbt: Isn't this equivalent to what we had before? If it is not Some it will not perform that check and wait for the next iteration?

crates/rbuilder/src/live_builder/order_input/mod.rs
```diff
@@ -334,33 +334,41 @@ where
     let handle = tokio::spawn(async move {
```
dvush, line 334: Usually this jobs need to be spawned as blocking and not as async tasks and its time consuming and blocks executor.

dvush: Its probably fine here as this job happens between slots.

dvush (approved):

crates/rbuilder/src/live_builder/order_input/mod.rs
ZanCorDX: I haven't tried it. Claude says None will not match and will not stop the loop so it may not behave exactly as before.
You get the exact behavior as before (stop con channel end) if you match header_res = header_receiver.recv() without the "Some" and then check match for Some (current code)/None (end)

ZanCorDX (approved):

PR #352, opened 2025-01-07
Add flashblocks payload builder

ferranbt:
## 📝 Summary

This PR introduces a payload generator for Reth that does not create multiple block building blocks but only spawns a single process.

## 💡 Motivation and Context

---

* [ ] Run `make lint`
* [ ] Run `make test`


ferranbt:
I believe the bench is going to fail for this PR because it compares execution at two different branches and when it checks out develop branch it keeps the new `op-rbuilder2` package and tries to run it without the optimism flag.

ZanCorDX (approved):

PR #360, opened 2025-01-15
Backtesting fixed

ZanCorDX:
## 📝 Summary

For some reason github allowed to merge https://github.com/flashbots/rbuilder/pull/344 with errors!
Fixed those errors and also disabled roothash calculation on backtesting.

## 💡 Motivation and Context

I was threatened by the the other programmers.

* [X] Run `make lint`
* [X] Run `make test`

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -7845,6 +7845,7 @@ dependencies = [
  "rbuilder",
  "reth-primitives",
  "reth-provider",
+ "reth-transaction-pool",
  "tokio",
  "tokio-util",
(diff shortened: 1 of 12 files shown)


ryanschneider (approved):
looks good to unblock us and get building working again, but I'm not a huge fan of the `skip_root_hash` argument for two reasons:

- it's "inverted" so the correct production usage is `create_provider_factory(false)` which feels odd.
- I feel like a separate `create_backtest_provider_factory` method might be more clear.

PR #361, opened 2025-01-15
add builder tx to new op-rbuilder

avalonche:
## 💡 Motivation and Context

---

* [ ] Run `make lint`
* [ ] Run `make test`


crates/op-rbuilder/Cargo.toml
```diff
@@ -1,9 +1,12 @@
 [package]
-name = "op-rbuilder2"
+name = "op-rbuilder"
 version = "0.1.0"
 edition = "2021"
 
 [dependencies]
+# workspace
+rbuilder = { path = "../rbuilder", features = ["optimism"] }
```
ferranbt: Please do not depend on rbuilder for now on this. It is okay to copy/paste code. I want to limit the number of moving parts right now. If we were to depend on something from rbuilder lets do a shared crate and depend there.

crates/op-rbuilder/src/args.rs
```diff
@@ -0,0 +1,111 @@
...
+/// Parameters for rollup configuration
+#[derive(Debug, Clone, Default, PartialEq, Eq, clap::Args)]
+#[command(next_help_heading = "Rollup")]
+pub struct OpRbuilderArgs {
+    /// Rollup configuration
+    #[command(flatten)]
+    pub rollup_args: RollupArgs,
+    /// Builder secret key for signing last transaction in block
+    #[arg(long = "rollup.builder-secret-key")]
+    pub builder_secret_key: Option<EnvOrValue<String>>,
```
ferranbt: Since we are using flags already, we do not need to use the `env:` pattern from the rbuilder. It adds an extra layer of indirection that is not necessary. We can use env variables in clap [like this](https://rust.code-maven.com/slides/rust/clap-and-environment-variables.html) and simplify this code.

crates/op-rbuilder/src/args.rs
```diff
@@ -0,0 +1,111 @@
...
+}
+
+impl OpRbuilderArgs {
+    pub fn builder_secret_key(&self) -> eyre::Result<Option<Signer>> {
```
ferranbt: I would prefer if we do not add this function and deserialize directly on the config file like [this](https://github.com/flashbots/rbuilder/pull/326).

ferranbt (changes requested):
A few things.

crates/op-rbuilder/Cargo.toml
avalonche: I'm thinking of a shared crate but imported due to time constraints 👍 I'll copy paste over the code

crates/op-rbuilder/src/args.rs
avalonche: why does the rbuilder use the env: pattern?

ferranbt (commented):
Are all the builder changes necessary now?

ZanCorDX (changes requested):
I don't like that you copy/pasted tx_signer.rs.
Why?
Unless imposible please correct it.

avalonche:
@ZanCorDX see https://github.com/flashbots/rbuilder/pull/361#discussion_r1918222057

ZanCorDX (approved):

ferranbt (approved):

PR #373, opened 2025-01-23
New relay definition parameters to split submit and get slot info roles.

ZanCorDX:
## 📝 Summary

The role of MevBoostRelay is now fulfilled by MevBoostRelayBidSubmitter or MevBoostRelaySlotInfoProvider.
This allows the use of some relays only to get slot info and others to bid.

## 💡 Motivation and Context

This will allow some new testing mode using a new relay mode "fake".
The fake relay will receive bids but does not implement any validation registration info.

---

* [X] Run `make lint`
* [X] Run `make test`

### DIFF
--- a/crates/rbuilder/src/primitives/mev_boost.rs
+++ b/crates/rbuilder/src/primitives/mev_boost.rs
@@ -1,5 +1,5 @@
 use crate::mev_boost::{
-    RelayClient, RelayError, SubmitBlockErr, SubmitBlockRequest, ValidatorSlotData,
+    RelayClient, RelayError, SubmitBlockErr, SubmitBlockRequestWithMetadata, ValidatorSlotData,
 };
 use governor::{DefaultDirectRateLimiter, Quota, RateLimiter};
 use serde::{Deserialize, Deserializer};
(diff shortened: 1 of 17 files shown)


crates/rbuilder/src/primitives/mev_boost.rs
```diff
@@ -1,31 +1,77 @@
...
+#[derive(Debug, Deserialize, Clone, PartialEq, Eq, Default)]
+pub enum RelayMode {
+    /// Submits bids, gets slot info. No extra headers on bidding.
+    #[serde(rename = "full")]
+    Full,
+    /// Only gets slot info.
+    #[serde(rename = "slot_info")]
+    #[default]
+    GetSlotInfoOnly,
+    /// Submits bids with extra headers. Does not used to get slot info.
```
ryanschneider:
```suggestion
    /// Submits bids with extra headers. Is not used to get slot info.
```

ryanschneider (approved):
lgtm!

metachris:
Is this backwards compatible, or does it require the `mode` setting now? I think it should probably stay backwards compatible 🤔 

Instead of `mode=fake`, how about naming it `mode=test`?

metachris:
Would also be good to add the modes to the documentation.

ZanCorDX:
> Would also be good to add the modes to the documentation.

Agree,  we have missing a good full cfg manual.
I'll make a new PR for that after this.

PR #376, opened 2025-01-24
Block gas limit calculation bugfix

### LINKED ISSUE #375: Block gas limit calculation has a bug
While testing the upcoming "remote state provider" I noticed that, since upgrading the `reth` dependency to v1.1.5, I've been getting a lot of errors regarding block gas limit when validating the `rbuilder` block using the `dry_run` flag. 

I'm attaching the screenshot:
Image

The "green box" marks passed validation, and the red box failed.

mralj:
## 📝 Summary

The issue for calculating the block gas limit has been fixed by simply deleting  "a double call" to calculate the gas limit (what's more wrong params were passed to this second/unnecessary call).

I have also removed the `calc_gas_limit` from `utils/mod.rs` as it seems unnecessary since the `rbuilder` depends on `alloy` where this calculation is already implemented.
I have left tests for gas limit calculations, but I would argue that they are redundant (since, if this change is accepted, we'll be depending on `alloy` for gas limit calculation)

## 💡 Motivation and Context

Fixes #375 

---

* [x] Run `make lint`
* [x] Run `make test`

### DIFF
--- a/crates/rbuilder/src/building/mod.rs
+++ b/crates/rbuilder/src/building/mod.rs
@@ -19,13 +19,19 @@ use crate::{
     primitives::{Order, OrderId, SimValue, SimulatedOrder, TransactionSignedEcRecoveredWithBlobs},
     provider::RootHasher,
     roothash::RootHashError,
-    utils::{a2r_withdrawal, calc_gas_limit, timestamp_as_u64, Signer},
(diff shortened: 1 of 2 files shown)


ZanCorDX (approved):

ryanschneider (approved):
Thanks for catching this @mralj!

PR #381, opened 2025-01-28
Add integration test for op-rbuilder

ferranbt:
This PR introduces an integration test framework for op-rbuilder that enables comprehensive end-to-end testing. The framework allows us to:

- Spin up a local op-rbuilder node with configurable parameters
- Test block production and chain advancement
- Capture and analyze logs for debugging
- Clean up resources automatically after tests complete

Key changes:

- Added IntegrationFramework to manage test services lifecycle
- Created test utilities for genesis file generation and block production
- Added initial test case verifying block production
- Configured CI workflow to run integration tests and the logs are saved as artifacts

To run the integration test run:

```
$ cargo run -p op-rbuilder --bin tester --features optimism -- genesis --output genesis.json
$ cargo build -p op-rbuilder --bin op-rbuilder --features optimism
$ cargo test --package op-rbuilder --lib --features optimism,integration -- integration::integration_test::tests
```

This framework is based on the testing scripts we currently use for L1 testing. A follow-up task will be to extract this into a separate crate that can be shared between L1 and L2 testing, standardizing our integration testing approach across the codebase.


crates/op-rbuilder/src/integration/mod.rs
```diff
@@ -0,0 +1,182 @@
...
+    io::prelude::*,
+    path::PathBuf,
+    process::{Child, Command},
+    time::{Duration, SystemTime},
+};
+use time::{format_description, OffsetDateTime};
+use tokio::time::sleep;
+
+/// Default JWT token for testing purposes
+pub const DEFAULT_JWT_TOKEN: &str =
```
avalonche, line 15: can move to a common crate for this constant

crates/op-rbuilder/src/integration/mod.rs
```diff
@@ -0,0 +1,182 @@
...
+    timeout: Duration,
+) -> Result<(), IntegrationError> {
+    let start = std::time::Instant::now();
+
+    loop {
+        if start.elapsed() > timeout {
+            return Err(IntegrationError::SpawnError);
+        }
+
+        let mut file = File::open(log_path).map_err(|_| IntegrationError::LogError)?;
```
avalonche, line 62: are there concurrency issues with polling logs and services writing to the logs?

ferranbt: That concurrency is handled at the os level. Worst case would be that you are reading something that has not been written yet but since it is used only for readiness checks it should be fine.

crates/op-rbuilder/src/integration/mod.rs
ferranbt: Why do you feel we need that? The constant is only used in the context of that integration package. Note that eventually the integration folder will be a crate in itself to be shared with L1.

avalonche (approved):

crates/op-rbuilder/src/integration/mod.rs
avalonche: I thought it was reused [here](https://github.com/flashbots/rbuilder/pull/381/files#diff-96be37b58b18ff7912d30856e6cc8e8dcdc160201ad8cc3ea7d4f35ccd0fa3b0R43-R45). Just seeing the same jwt reused but not a blocker

PR #385, opened 2025-01-29
Expand integration test for blocklist

ferranbt:
This PR expands the integration test to cover blocklist functionality end-to-end. The test sends transactions from and to blocklisted accounts to verify the blocklist works as expected. All scenarios are tested within the same test function since the integration framework doesn't currently support concurrent test execution (this should be fixed in the future but requires deeper architectural changes).

Added debug logging when transactions are rejected due to blocklist, allowing the test to verify correct behavior by checking log output.

### DIFF
--- a/crates/rbuilder/src/building/order_commit.rs
+++ b/crates/rbuilder/src/building/order_commit.rs
@@ -394,12 +394,25 @@ impl<'a, 'b, Tracer: SimulationTracer> PartialBlockFork<'a, 'b, Tracer> {
 
         let mut db = self.state.new_db_ref();
         let tx = &tx_with_blobs.internal_tx_unsecure();
-        if ctx.blocklist.contains(&tx.signer())
-            || tx
-                .to()
-                .map(|to| ctx.blocklist.contains(&to))
-                .unwrap_or(false)
+        // Check if sender is blocklisted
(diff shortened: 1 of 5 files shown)


crates/rbuilder/src/building/order_commit.rs
```diff
@@ -394,12 +394,25 @@ impl<'a, 'b, Tracer: SimulationTracer> PartialBlockFork<'a, 'b, Tracer> {
...
+        // Check if recipient is blocklisted
+        if tx
+            .to()
+            .map(|to| ctx.blocklist.contains(&to))
+            .unwrap_or(false)
         {
+            tracing::debug!(
+                "Transaction rejected - recipient {} is blocklisted",
+                tx.to().unwrap()
+            );
```
ZanCorDX, line 415: I'm not sure we need this traces it's a super common valid case.

ZanCorDX (changes requested):
The implementation adds really anoying traces just to be able to test this.
Checking the logs for a test is super unstable (strings change a lot). In case you have to, you should refactor the string constants to a single point.

ZanCorDX:
already covered on https://github.com/flashbots/rbuilder/blob/develop/crates/rbuilder/src/building/testing/bundle_tests/mod.rs

PR #387, opened 2025-01-29
deps: reth v1.1.6

efbig:
## 💡 Motivation and Context

Migration to reth deps `v1.1.6`. Supersedes https://github.com/flashbots/rbuilder/pull/370.

---

* [ ] Run `make lint`
* [ ] Run `make test`

### DIFF
--- a/crates/reth-rbuilder/src/main.rs
+++ b/crates/reth-rbuilder/src/main.rs
@@ -20,13 +20,13 @@ use reth_node_builder::{
 };
 use reth_node_ethereum::{node::EthereumAddOns, EthereumNode};
 use reth_provider::{
-    providers::{BlockchainProvider, BlockchainProvider2},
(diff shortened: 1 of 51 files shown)


crates/reth-rbuilder/src/main.rs
```diff
@@ -69,49 +62,31 @@ fn main() {
...
-                        .with_add_ons::<EthereumAddOns<_>>(Default::default())
-                        .on_node_started(move |node| {
-                            spawn_rbuilder(node.provider().clone(), node.pool().clone(), extra_args.rbuilder_config);
-                            Ok(())
-                        })
-                        .launch().await?;
-                    handle.node_exit_future.await
-                }
-            }
+            let engine_tree_config = TreeConfig::default()
```
gakonst: Heads up should be able to do this simpler now, given we did this for OP here https://github.com/paradigmxyz/reth/pull/13900 I think?

PR #389, opened 2025-01-29
Dynamic load of blocklist from url

ZanCorDX:
## 📝 Summary

Now we have a new config field blocklist which can be:

a file name and behaves as before (loads from the file)
a url and downloads the file from the web. It also updates periodically the list. If it can't update and the blocklist get's too old (configurable via blocklist_url_max_age_hours (default 24)) it closes the builder.
We still support the old blocklist_file_path.
The change involved replacing several blocklists for a new trait BlockListProvider.

## 💡 Motivation and Context

More robustness.
---

* [X] Run `make lint`
* [X] Run `make test`

### DIFF
--- a/crates/rbuilder/src/live_builder/block_list_provider.rs
+++ b/crates/rbuilder/src/live_builder/block_list_provider.rs
@@ -1,0 +1,193 @@
+use ahash::HashSet;
+use revm_primitives::Address;
+use std::{
+    fs::read_to_string,
+    path::PathBuf,
+    sync::{Arc, Mutex, MutexGuard},
+    time::Duration,
+};
+use time::OffsetDateTime;
+use tokio_util::sync::CancellationToken;
+use tracing::{error, info};
+use url::Url;
+
+pub type BlockList = HashSet<Address>;
+
+#[derive(Debug, thiserror::Error)]
+pub enum Error {
+    #[error("Unable to load initial list")]
+    UnableToLoadInitialList,
+    #[error("Unable to update list")]
+    UnableToUpdateList,
+}
+
+const MIN_ITEMS_ON_BLOCK_LIST: usize = 1;
+
+/// Basic validation on a block list.
+/// Returns if the list looks good.
+fn validate_list(blocklist: &[Address]) -> bool {
+    blocklist.len() >= MIN_ITEMS_ON_BLOCK_LIST
+}
+
+/// Abstraction to get and update the builder's blocklist.
+pub trait BlockListProvider: std::fmt::Debug + Sync + Send {
+    /// Gets the a copy of the last list. Fails if it's too old.
+    fn get_blocklist(&self) -> Result<BlockList, Error>;
+    /// Checks a single address in the current list. Fails if it's too old.
+    fn current_list_contains(&self, address: &Address) -> Result<bool, Error>;
+}
+
+/// BlockListProvider that always returns Ok(empty list)
+#[derive(Debug)]
+pub struct NullBlockListProvider {}
+
+impl BlockListProvider for NullBlockListProvider {
+    fn get_blocklist(&self) -> Result<BlockList, Error> {
+        Ok(BlockList::default())
+    }
+
+    fn current_list_contains(&self, _address: &Address) -> Result<bool, Error> {
+        Ok(false)
+    }
+}
+
+#[derive(Debug)]
+struct BlockListWithTimestamp {
+    pub block_list: BlockList,
+    pub timestamp: OffsetDateTime,
+}
+
+impl BlockListWithTimestamp {
+    fn new(block_list: BlockList) -> Self {
+        Self {
+            block_list,
+            timestamp: OffsetDateTime::now_utc(),
+        }
+    }
+}
+
+const TIMES_TO_UPDATE_PER_MAX_AGE: f32 = 10.0;
+/// BlockListProvider that downloads the list from a url and tries to keep it up to date.
+/// If it fails to update on time it gives an error.
+#[derive(Debug)]
+pub struct HttpBlockListProvider {
+    max_allowed_age: Duration,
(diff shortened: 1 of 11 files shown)


crates/rbuilder/src/live_builder/block_list_provider.rs
```diff
@@ -0,0 +1,193 @@
...
+            return Err(Error::UnableToLoadInitialList);
+        }
+        Ok(Self {
+            block_list: Mutex::new(blocklist.into_iter().collect()),
+        })
+    }
+}
+
+impl BlockListProvider for StaticFileBlockListProvider {
+    /// Just gets the last version and checks the age.
```
dvush: thats invalid comment

crates/rbuilder/src/live_builder/block_list_provider.rs
```diff
@@ -0,0 +1,193 @@
...
+        Ok(last_updated_list.block_list.contains(address))
+    }
+}
+
+/// BlockListProvider that opens a file. Useful for backtesting and static scenarios.
+/// Can only fail on creation.
+/// @Pending upgrade the HttpBlockListProvider to allow to plugin the reader and have a generic updatable source for http/file
+#[derive(Debug)]
+pub struct StaticFileBlockListProvider {
+    block_list: Mutex<BlockList>,
```
dvush: Why mutex is needed here?

crates/rbuilder/src/live_builder/payload_events/mod.rs
```diff
@@ -150,11 +150,21 @@ impl MevBoostSlotDataGenerator {
...
+                    &mev_boost_slot_data,
+                    self.blocklist_provider.as_ref(),
+                ) {
+                    Ok(can_build) => {
+                        if !can_build {
+                            continue;
+                        }
+                    }
+                    Err(_) => {
+                        // Blocklist errors are FATAL
```
dvush, line 163: Wait, what is that? Anyone can set payload fee recipient to a block list address. We don't need to kill the builder because of that, we just need to skip this slot.

crates/rbuilder/src/live_builder/payload_events/mod.rs
```diff
@@ -190,17 +200,18 @@ impl SlotSource for MevBoostSlotDataGenerator {
     }
 }
 
+/// true->build
+/// false->don't build
+/// Error crisis, close.
```
dvush, line 205: There is no crisis (see comment above)

dvush (changes requested):

crates/rbuilder/src/live_builder/block_list_provider.rs
ZanCorDX: you are right! Copy/pasted it!

crates/rbuilder/src/live_builder/block_list_provider.rs
ZanCorDX: you are right! Came from a copy/paste!

crates/rbuilder/src/live_builder/payload_events/mod.rs
ZanCorDX: This applies also to the other comment: check_slot_data_for_blocklist return changed if we can't get a good blocklist (it's too old) we return an error and that's critical.
If we have a good blocklist we get Ok(false/true) which is fine  models the problem you mentioned ( fee recipient to a block list address)

dvush (approved):

crates/rbuilder/src/live_builder/block_list_provider.rs
```diff
@@ -0,0 +1,192 @@
...
+fn validate_list(blocklist: &[Address]) -> bool {
+    blocklist.len() >= MIN_ITEMS_ON_BLOCK_LIST
+}
+
+/// Abstraction to get and update the builder's blocklist.
+pub trait BlockListProvider: std::fmt::Debug + Sync + Send {
+    /// Gets the a copy of the last list. Fails if it's too old.
+    fn get_blocklist(&self) -> Result<BlockList, Error>;
+    /// Checks a single address in the current list. Fails if it's too old.
+    fn current_list_contains(&self, address: &Address) -> Result<bool, Error>;
```
ferranbt, line 37: Why do you want to have both a function to return whether the list contains an address and then to return a blockList reference that you can use to check whether the address is on the block list again?

crates/rbuilder/src/live_builder/block_list_provider.rs
```diff
@@ -0,0 +1,192 @@
...
+const MIN_ITEMS_ON_BLOCK_LIST: usize = 1;
+
+/// Basic validation on a block list.
+/// Returns if the list looks good.
+fn validate_list(blocklist: &[Address]) -> bool {
+    blocklist.len() >= MIN_ITEMS_ON_BLOCK_LIST
+}
+
+/// Abstraction to get and update the builder's blocklist.
+pub trait BlockListProvider: std::fmt::Debug + Sync + Send {
```
ferranbt, line 33: This trait is behaving like a factory method. Why do you want to have a trait then?
A provider pattern would give access to the data without transferring ownership.

crates/rbuilder/src/live_builder/block_list_provider.rs
ZanCorDX: To avoid making a whole copy in some contexts.

crates/rbuilder/src/live_builder/block_list_provider.rs
ZanCorDX: Yes the trait can be seen as similar as the factory pattern so I don't understan de question.

ferranbt: You called it provider. Provider and factory are different things to me.

PR #396, opened 2025-01-30
Blocklist integration tests

ZanCorDX:
## 📝 Summary

- Upgraded main (from @ferranbt 's PR https://github.com/flashbots/rbuilder/pull/385) test to check exclusion of a tx from/to a blocklisted address
- Added an integration test checking the builder closes if it can't have an up-to-date blocklist.

* [X] Run `make lint`
* [X] Run `make test`

### DIFF
--- a/.github/workflows/checks.yaml
+++ b/.github/workflows/checks.yaml
@@ -111,7 +111,7 @@ jobs:
         run: builder-playground &
 
       - name: Run integration tests with flags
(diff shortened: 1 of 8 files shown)


dvush (commented):
There is also a case that we test in our mev test contract when blocklisted call happens internally.

ferranbt (approved):

PR #398, opened 2025-01-31
Metrics improvements.

dvush:
## 📝 Summary

* Reset histograms periodically, otherwise recent changes do not affect them in the long running process.
* Make fill_time less biased as we now record every fill instead of the finalized block fill (not every block is finalized)
* Add counter for received orderpool commands by type (orderpool_commands_received)
* Add histogram of order simualtion time (order_simulation_time)
* Remove flawed e2e latency metric.
* Remove gas per second metric from simulations.
* Adds E2E global latency metrics

Global E2E latency metrics are constructed in a way to cover the whole path of the bundle without gaps.
For this we record events such as: bundle was received, bundle was simulated, bundle is considered for inclusion by this builder, etc. This is similar to tracing but expose aggregated metrics for each "bundle history".

There are multiple steps that are needed to prevent bias: only consider bundle that arrive when we actively submit blocks, exclude bundles that can't be landed because of nonce or other bundle with the same nonce, etc.

One of the findings from this is that: `2*fill time + finalize time` is a good proxy for p99 latency. (2x fill time because while builder building a block it does not take new bundle into account) and there are no major gaps in the builder that can't be explained.

* [x] Run `make lint`
* [x] Run `make test`

### DIFF
--- a/crates/rbuilder/src/building/builders/mock_block_building_helper.rs
+++ b/crates/rbuilder/src/building/builders/mock_block_building_helper.rs
@@ -1,5 +1,3 @@
-use std::fmt::Debug;
-
 use crate::live_builder::simulation::SimulatedOrderCommand;
 use crate::provider::RootHasher;
 use crate::roothash::RootHashError;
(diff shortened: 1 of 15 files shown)


crates/rbuilder/src/building/builders/mock_block_building_helper.rs
```diff
@@ -1,3 +1,5 @@
+use std::fmt::Debug;
+
```
ZanCorDX, line 2: Why?

ZanCorDX (approved):

PR #401, opened 2025-02-04
Proper replacements

ZanCorDX:
## 📝 Summary

-  Parallel sealer killed (config parameter max_concurrent_seals gone)
- Removed every check for increasing TBV
- Removed every check for increasing competition bid
- Added a UnfinishedBlockBuildingSinkMuxer at the output of the BlockBuildingAlgorithms to choose the best block among the last block from each algo
- Added cancellations in relay submiton (/relay/v1/builder/blocks?cancellations=1)
## 💡 Motivation and Context

Good order replacement is critical!

* [X] Run `make lint`
* [X] Run `make test`


dvush:
@ZanCorDX I don't know if relay respect it but we need to submit with a url param to enable this https://github.com/flashbots/geth-builder-old-unused/blob/b6a6dd336988d3f866a6c611a6505885966ba189/builder/relay.go#L142

ZanCorDX:
> @ZanCorDX I don't know if relay respect it but we need to submit with a url param to enable this https://github.com/flashbots/geth-builder-old-unused/blob/b6a6dd336988d3f866a6c611a6505885966ba189/builder/relay.go#L142

The code will never cancel, it replaces. From what I read on the relay spec (https://flashbots.github.io/relay-specs/#/Builder/submitBlock) it should work:
```
Any new submission by a builder will overwrite a previous one by the same builder_pubkey, even if it is less profitable.
```

crates/rbuilder/src/primitives/mev_boost.rs
```diff
@@ -132,6 +135,7 @@ impl MevBoostRelayBidSubmitter {
             optimistic: config.optimistic,
             submission_rate_limiter,
             test_relay,
+            cancellations: true,
```
dvush, line 138: relays (our implementation) will reject block with this flag if its not configured for replacements. Thats likely to be non-issue but I am noting here so when we test we actually need to check grafana for relay errors.

crates/rbuilder/src/building/built_block_trace.rs
```diff
@@ -164,4 +164,14 @@ impl BuiltBlockTrace {
 
         Ok(())
     }
+
+    pub fn transaction_hashes(&self) -> Vec<TxHash> {
+        let mut hashes = Vec::new();
```
dvush: nit: you can preallocate at least self.included_orders + some small number

crates/rbuilder/src/live_builder/block_output/relay_submit.rs
```diff
@@ -168,16 +169,18 @@ async fn run_submit_to_relays_job(
         (normal_relays, optimistic_relays)
     };
 
-    let mut last_bid_value = U256::from(0);
+    let mut last_bid_hash = None;
```
dvush, line 172: nit, to make code simpler and remove this option you can set it to B256::ZERO

crates/rbuilder/src/live_builder/building/unfinished_block_building_sink_muxer.rs
```diff
@@ -0,0 +1,243 @@
...
+        }
+    }
+
+    /// Given a new BiddableUnfinishedBlock updates the best per algo and return the best one if
+    /// changes from last time.
+    fn update_best_block(
+        &mut self,
+        biddable_block: BiddableUnfinishedBlock,
+    ) -> Option<BiddableUnfinishedBlock> {
+        self.algorithm_to_last_block.insert(
```
dvush, line 73: nit: allocating string every time here is not necessary as its going to be filled and replaced

crates/rbuilder/src/live_builder/building/unfinished_block_building_sink_muxer.rs
```diff
@@ -0,0 +1,243 @@
...
+            biddable_block.block().builder_name().to_string(),
+            biddable_block,
+        );
+        let (_, best_biddable_block) = self
+            .algorithm_to_last_block
+            .iter()
```
dvush: nit: there is a values() iter for hash map values

crates/rbuilder/src/live_builder/building/unfinished_block_building_sink_muxer.rs
```diff
@@ -0,0 +1,243 @@
...
+    }
+}
+
+/// Computes the best block among all the algos.
+struct BestBlockState {
+    /// For each algo name we store the last block so on each update we can look for the best block among all algos.
+    /// We should replace String for "real" type identifying the algo.
+    algorithm_to_last_block: HashMap<String, BiddableUnfinishedBlock>,
+    /// Works as an id for the last block so we don't send the same twice since the same block could be generated by different algos.
+    last_sent_block_txs: Option<Vec<TxHash>>,
```
dvush: nit: I am not a fan on copying 300*32 bytes in the array just to see if txs are the same + comparing it. I would use u64 computed from &B256 hash using ahash hasher.

crates/rbuilder/src/building/built_block_trace.rs
dvush: see nit below where I propose to use u64 computed as ahash from &B256 txs

crates/rbuilder/src/live_builder/building/unfinished_block_building_sink_muxer.rs
dvush: also, this option can be probably removed for just empty vec

crates/rbuilder/src/primitives/mev_boost.rs
dvush: if this happens we would need to set this value per relay in config
also, we can just add this to relay config already just in case

dvush:
also, would it make sense to have this as a feature flag? I am not against this but there may be a problem if it does not work for some reason and we would need to quickly update builder net binary instead of config

crates/rbuilder/src/building/built_block_trace.rs
dvush: or you can XOR last limb of tx hashes into u64 since tx hashes are already hashes

crates/rbuilder/src/primitives/mev_boost.rs
ZanCorDX: The thing is that this is a deal breaker, without cancelation it should fail. If add a flag to the cfg and somebody sets it on false by mistake it would be really bad.

ZanCorDX:
> also, would it make sense to have this as a feature flag? I am not against this but there may be a problem if it does not work for some reason and we would need to quickly update builder net binary instead of config
Mmmm... that would take me a while. Maybe for the future?

ZanCorDX:
Changed it to use a hash from tx hashs to avoid coping stuff.

crates/rbuilder/src/live_builder/building/unfinished_block_building_sink_muxer.rs
```diff
@@ -54,14 +53,14 @@ struct BestBlockState {
     /// We should replace String for "real" type identifying the algo.
     algorithm_to_last_block: HashMap<String, BiddableUnfinishedBlock>,
     /// Works as an id for the last block so we don't send the same twice since the same block could be generated by different algos.
-    last_sent_block_txs: Option<Vec<TxHash>>,
+    last_sent_block_hash: Option<u64>,
```
dvush, line 56: why not 0?

ZanCorDX: When possible, I don't like to encode special cases in numbers, I think it hides behavior. 
In my life I had way too many intergers ending up on -1/0/int.max/int.min by mistake causing unpredicted bugs.

dvush (approved):

PR #404, opened 2025-02-05
Root hash thread pool, new trie benches.

dvush:
## 📝 Summary

* adds a new, configurable thread pool used for root hash with custom thread names
* refactors sparse trie benches and tests to use gzip files
* adds more benches for real tries with focus on p99 latency

## 💡 Motivation and Context

I was experimenting with thread affinity and improved eth sparse trie along the way. Affinity does not appear to give boost in real builder so its not here. We should revisit it in the future.

* [ ] Run `make lint`
* [ ] Run `make test`


ZanCorDX (changes requested):
StateProviderFactory::root_hasher should not have "thread_pool: Option<RootHashThreadPool>", since it's for a specific implementation.
It should be in StateProviderFactoryFromProviderFactory/StateProviderFactoryFromRethProvider the same way we hide the root_hash_config from the roothash users (maybe StateProviderFactoryFromRethProvider/StateProviderFactoryFromProviderFactory need a refactoring?).

dvush:
good point, I moved it to root hash config (now called context) since its begins from config file and its being copied from there.

ZanCorDX (approved):

PR #405, opened 2025-02-06
correctly parse eth_sendBundle payload when sent as object

zeroXbrock:
## 📝 Summary

parse `eth_sendBundle` params as object if sent as object, otherwise parse as single-element array (prev behavior).

## 💡 Motivation and Context

if params are sent as an object (as is done by [alloy](https://github.com/alloy-rs)), `eth_sendBundle` fails with an "invalid payload" error.

---

* [x] Run `make lint`
* [x] Run `make test`
~~* [ ] Added tests (if applicable)~~

### DIFF
--- a/crates/rbuilder/src/live_builder/order_input/rpc_server.rs
+++ b/crates/rbuilder/src/live_builder/order_input/rpc_server.rs
@@ -46,10 +46,15 @@ pub async fn start_server_accepting_bundles(
     module.register_async_method("eth_sendBundle", move |params, _| {
         let results = results_clone.clone();
         async move {
-	    let received_at = OffsetDateTime::now_utc();
-
+    	    let received_at = OffsetDateTime::now_utc();
             let start = Instant::now();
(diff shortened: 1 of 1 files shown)


ZanCorDX:
Acording to the spec (https://docs.flashbots.net/flashbots-auction/advanced/rpc-endpoint), it doesn't seem to explicitly support your case. Can't you solve this in the sender side?

dvush:
I think its better to fix this on the sender side because:
* this also consistent with other infra, proxies, filters that are used in front of the builder (such as tdx-orderflow-proxy, etc). This means that changing this in the builders will not be a full solution and can bring more confusion because sending this bundle format to rpc will not work
* this seems to be a standard how it works across builders / rpcs that I've checked right now

Given the points above I think we should not do that for consistency. 

Please, reopen the issue if there are more arguments for this change.

dvush:
To fix this on the sender just set it as an array of 1 element explicitly if your library always assumes objects unless its array.

PR #410, opened 2025-02-10
Validate rbuilder with op-reth

ferranbt:
## 📝 Summary

This PR extends the `op-rbuilder` integration test to use an off-the-shelf op-node to validate the blocks built. It was already possible to do as part of the engineAPI mocker command. This PR enables that behaviour on the integration test too.

## 💡 Motivation and Context

---

* [x] Run `make lint`
* [x] Run `make test`

### DIFF
--- a/.github/workflows/checks.yaml
+++ b/.github/workflows/checks.yaml
@@ -165,6 +165,9 @@ jobs:
           echo "SCCACHE_GHA_ENABLED=true" >> $GITHUB_ENV
           echo "RUSTC_WRAPPER=sccache" >> $GITHUB_ENV
 
+      - name: Download op-reth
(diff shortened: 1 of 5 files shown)


ZanCorDX (approved):

.github/workflows/checks.yaml
```diff
@@ -165,6 +165,9 @@ jobs:
           echo "SCCACHE_GHA_ENABLED=true" >> $GITHUB_ENV
           echo "RUSTC_WRAPPER=sccache" >> $GITHUB_ENV
 
+      - name: Download op-reth
```
ferranbt, line 168: I forgot to add this binary to $PATH. Will do in the afternoon.

PR #416, opened 2025-02-10
Use payload attributes timestamp to calculate op payload deadline

### LINKED ISSUE #413: Remove hardcoded deadline from Op payload generator
The BlockPayloadGenerator has a hardcoded deadline of 2 seconds. We should use the `attributes.timestamp()` field to set the deadline instead.

ferranbt:
## 📝 Summary

This PR removes the hardcoded deadline for the Op payload builder of 2 seconds and uses the payload attributes timestmap to calculate it.

Closes #413 

## 💡 Motivation and Context

---

* [ ] Run `make lint`
* [ ] Run `make test`

### DIFF
--- a/crates/op-rbuilder/src/generator.rs
+++ b/crates/op-rbuilder/src/generator.rs
@@ -12,6 +12,8 @@ use reth_payload_builder::PayloadJobGenerator;
 use reth_payload_builder::{KeepPayloadJobAlive, PayloadBuilderError, PayloadJob};
 use reth_payload_primitives::BuiltPayload;
 use std::sync::{Arc, Mutex};
+use std::time::SystemTime;
+use std::time::UNIX_EPOCH;
 use tokio::sync::oneshot;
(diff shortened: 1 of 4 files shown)


crates/op-rbuilder/src/generator.rs
```diff
@@ -366,6 +377,23 @@ impl<T: Clone> Default for BlockCell<T> {
...
+
+    // Safe subtraction that handles the case where timestamp is in the past
+    let duration_until = unix_timestamp_secs.saturating_sub(unix_now);
+
+    if duration_until == 0 {
+        // Enforce a minimum block time of 1 second by rounding up any duration less than 1 second
+        Duration::from_secs(1)
+    } else {
+        Duration::from_secs(duration_until)
+    }
```
ZanCorDX, line 394: shouldn't this be duration_until < 1 sec?

ZanCorDX (approved):

PR #422, opened 2025-02-11
remove broken benchmark CI

metachris:
## 📝 Summary

Benchmark in CI has been broken for a while:
- Metrics were inaccurate and gave the wrong picture
- Links stopped working a while ago

Because of these, nobody looks at them anymore since a while already. Better to remove them, and maybe re-add later on.

---

* [x] Run `make lint`
* [x] Run `make test`


sukoneck (approved):
Makes sense, appreciate the clean up. This was a good idea, let's come back to this at some point!

PR #423, opened 2025-02-11
qol(reth-rbuilder): default to `0` on engine persistence settings

rkrasiuk:
## 📝 Summary

Default to `0` on engine persistence settings because this is how it's supposed to be used anyway.


ryanschneider (approved):

ryanschneider:
So for added context, reth defaults to `2` for these values, but `rbuilder` should really default to `0` since it doesn't really work unless `0` even when in-process.

PR #424, opened 2025-02-12
Add cancellation token + fb-rb integration

ferranbt:
## 📝 Summary

This PR replaces the Cancellation bool behind a mutex from Reth with a tokio Cancellation token that we can cancel from different places.

## 💡 Motivation and Context

---

* [ ] Run `make lint`
* [ ] Run `make test`


crates/op-rbuilder/src/generator.rs
```diff
@@ -114,6 +122,26 @@ where
         &self,
         attributes: <Builder as PayloadBuilder<Pool, Client>>::Attributes,
     ) -> Result<Self::Job, PayloadBuilderError> {
+        let cancel_token = if self.ensure_only_one_payload {
+            // Cancel any existing payload first
+            {
+                let last_payload = self.last_payload.lock().unwrap();
+                if let Some(token) = &*last_payload {
```
dvush: thats nice, I did not know that this works (for matching mutex)

crates/op-rbuilder/src/generator.rs
```diff
@@ -114,6 +122,26 @@ where
         &self,
         attributes: <Builder as PayloadBuilder<Pool, Client>>::Attributes,
     ) -> Result<Self::Job, PayloadBuilderError> {
+        let cancel_token = if self.ensure_only_one_payload {
+            // Cancel any existing payload first
+            {
+                let last_payload = self.last_payload.lock().unwrap();
```
dvush, line 128: I wonder why you lock twice here? If you think that token can hold the lock its still not going to help since you hold it while cancelling

crates/op-rbuilder/src/generator.rs
```diff
@@ -68,6 +69,10 @@ pub struct BlockPayloadJobGenerator<Client, Pool, Tasks, Builder> {
     ///
     /// See [PayloadBuilder]
     builder: Builder,
+    /// Whether to ensure only one payload is being processed at a time
+    ensure_only_one_payload: bool,
+    /// The last payload being processed
+    last_payload: Arc<Mutex<Option<CancellationToken>>>,
```
dvush: nitpick: since you always cancel you can just have value here all the time without any option

ZanCorDX (approved):

crates/op-rbuilder/src/payload_builder.rs
```diff
@@ -52,24 +61,92 @@ pub struct OpPayloadBuilder<EvmConfig, Txs = ()> {
...
     }
+
+    /// Start the WebSocket server
+    pub async fn start_ws(&self, addr: &str) -> Result<(), Box<dyn std::error::Error>> {
+        let listener = TcpListener::bind(addr).await?;
+        let subscribers = self.subscribers.clone();
+
+        tokio::spawn(async move {
+            while let Ok((stream, _)) = listener.accept().await {
+                println!("Accepted websocket connection");
```
dvush: is println intended here instead of tracing?

dvush (approved):

PR #429, opened 2025-02-13
Reth 1.2.0

ZanCorDX:
## 📝 Summary

Lovely reth changes.

## 💡 Motivation and Context

Pectra

* [X] Run `make lint`
* [X] Run `make test`

### DIFF
--- a/crates/rbuilder/src/backtest/restore_landed_orders/resim_landed_block.rs
+++ b/crates/rbuilder/src/backtest/restore_landed_orders/resim_landed_block.rs
@@ -10,12 +10,12 @@ use ahash::{HashMap, HashSet};
 use alloy_primitives::{TxHash, B256, I256};
 use eyre::Context;
 use reth_chainspec::ChainSpec;
-use reth_primitives::{Receipt, TransactionSignedEcRecovered};
+use reth_primitives::{Receipt, Recovered, TransactionSigned};
 use std::sync::Arc;
 
 #[derive(Debug)]
 pub struct ExecutedTxs {
-    tx: TransactionSignedEcRecovered,
+    tx: Recovered<TransactionSigned>,
     pub receipt: Receipt,
     pub coinbase_profit: I256,
     pub conflicting_txs: Vec<(B256, Vec<SlotKey>)>,
@@ -23,7 +23,7 @@ pub struct ExecutedTxs {
 
 impl ExecutedTxs {
     pub fn hash(&self) -> TxHash {
-        self.tx.hash()
+        *self.tx.hash()
     }
 }
 
@@ -42,7 +42,7 @@ where
     let suggested_fee_recipient = find_suggested_fee_recipient(&onchain_block, &txs);
 
     let coinbase = onchain_block.header.beneficiary;
-    let parent_hash = onchain_block.header.parent_hash;
+    let parent_num_hash = onchain_block.header.parent_num_hash();
 
     let ctx = BlockBuildingContext::from_onchain_block(
         onchain_block,
@@ -52,7 +52,7 @@ where
         coinbase,
         suggested_fee_recipient,
         None,
-        Arc::from(provider.root_hasher(parent_hash)?),
+        Arc::from(provider.root_hasher(parent_num_hash)?),
     );
 
     let state_provider = provider.history_by_block_hash(ctx.attributes.parent)?;
--- a/crates/rbuilder/src/building/mod.rs
+++ b/crates/rbuilder/src/building/mod.rs
@@ -55,7 +55,7 @@ use revm::{
     db::states::bundle_state::BundleRetention,
     primitives::{BlobExcessGasAndPrice, BlockEnv, SpecId},
 };
-use revm_primitives::InvalidTransaction;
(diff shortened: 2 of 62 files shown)


crates/rbuilder/src/building/mod.rs
```diff
@@ -626,18 +622,10 @@ impl<Tracer: SimulationTracer> PartialBlock<Tracer> {
...
-                )
+                .post_block_withdrawal_requests_contract_call(db.as_mut(), &ctx.evm_env)
                 .map_err(|err| FinalizeError::Other(err.into()))?;
             let consolidation_requests = system_caller
-                .post_block_consolidation_requests_contract_call(
-                    db.as_mut(),
-                    &ctx.initialized_cfg,
-                    &ctx.block_env,
-                )
+                .post_block_consolidation_requests_contract_call(db.as_mut(), &ctx.evm_env)
```
ryanschneider, line 626: I'm pretty sure we need to call:

```
db.as_mut().merge_transitions(BundleRetention::Reverts);
```

somewhere in this block, e.g. before `Some(requests)` at line 641. h/t to Aresnii for spotting this on telegram!

crates/rbuilder/src/backtest/restore_landed_orders/resim_landed_block.rs
```diff
@@ -10,20 +10,20 @@ use ahash::{HashMap, HashSet};
...
+    tx: Recovered<TransactionSigned>,
     pub receipt: Receipt,
     pub coinbase_profit: I256,
     pub conflicting_txs: Vec<(B256, Vec<SlotKey>)>,
 }
 
 impl ExecutedTxs {
     pub fn hash(&self) -> TxHash {
-        self.tx.hash()
+        *self.tx.hash()
```
dvush, line 26: I like this new change, lets return &TxHash instead here?

crates/rbuilder/src/building/mod.rs
```diff
@@ -137,7 +134,7 @@ impl BlockBuildingContext {
                 },
             )
             .ok()?;
-        block_env.coinbase = signer.address;
+        evm_env.block_env.coinbase = signer.address;
```
ryanschneider, line 137: Do we need https://github.com/rkrasiuk/rbuilder/commit/d7b1637967dffc1f4f3ade5303aacc185963c5c3 /cc @rkrasiuk

crates/rbuilder/src/building/mod.rs
ryanschneider: With the refactor this is now called down on line 654 (since the db is no longer shadowed).

ryanschneider (commented):
overall looks good to me, but curious if we need @rkrasiuk's bloblgas fix from his draft?

dvush (changes requested):
this branch breaks root hash and produces incorrect blocks a lot

no merge until fixed

dvush (approved):
After adding this fix https://github.com/flashbots/rbuilder/pull/429/commits/69ee27be67c493887acb6151e58987613d0fefea I don't see errors anymore on staging, it also does not affect performance since this case only happens when reth syncs new block and we should cancel job anyway.

Since default root hash does not work correctly either (since the problem is in ConsistentDbView I've added this temporarily so we can merge this PR) https://github.com/flashbots/rbuilder/pull/429/commits/29cc0fca432d422241f4ff75fcbcb371e40e6bd1

crates/rbuilder/src/live_builder/base_config.rs
```diff
@@ -297,6 +298,10 @@ impl BaseConfig {
                 "root_hash_compare_sparse_trie can't be set without root_hash_use_sparse_trie"
             );
         }
+        // temporary guard until reth is fixed
+        if !self.root_hash_use_sparse_trie || self.root_hash_compare_sparse_trie {
+            eyre::bail!("root_hash_use_sparse_trie=true and root_hash_compare_sparse_trie=false must be set, otherwise node will produce incorrect blocks or confusing error messages. These settings are enforced temporarily because upstream parallel root hash implementation is not correct.")
```
ryanschneider, line 303: @sukoneck we'll need to make sure that we set:

```
root_hash_use_sparse_trie = true
root_hash_compare_sparse_trie=false
```

On the testnet rbuilders once this is merged and then probably revert once fixed.

ryanschneider (approved):
lgtm!  Two potential follow-ups, we will do these as separate PRs:

- Fix `reth-rbuilder` handling of conflicting arguments: https://github.com/flashbots/rbuilder/pull/434/commits/bb52c8f4a8035bcc2f1479c0bff06b0faad193db
- Determine if we still need this blob gas calc change: https://github.com/rkrasiuk/rbuilder/commit/d7b1637967dffc1f4f3ade5303aacc185963c5c3

crates/rbuilder/src/live_builder/base_config.rs
dvush: This is our production configuration. I don't think we run even one builder without sparse trie. Its mostly for some other setups that may not use it for some reason.

PR #436, opened 2025-02-19
 New witness calculation and DA config

SozinM:
## 📝 Summary

New witness calculation algo.
DA config.
Reserve space for builder_tx in block_da_size.

Note:
If da_config.max_da_tx_size is too small, execute function of OpBuilder will panic, because it's improssible to insert builder transaction.
estimate_builder_tx_da_size may be a little bit heavy, maybe we could precompute tx size for builder_tx and reuse it

## 💡 Motivation and Context

---

## ✅ I have completed the following steps:

* [x] Run `make lint`
* [x] Run `make test`
* [ ] Added tests (if applicable)

PR #439, opened 2025-02-20
Remove task trait

SozinM:
## 📝 Summary
Could be merged after https://github.com/flashbots/rbuilder/pull/436
Closes https://github.com/flashbots/op-rbuilder/issues/23
## 💡 Motivation and Context

---

* [ ] Run `make lint`
* [ ] Run `make test`

### DIFF
--- a/crates/op-rbuilder/src/generator.rs
+++ b/crates/op-rbuilder/src/generator.rs
@@ -1,7 +1,7 @@
 use futures_util::Future;
 use futures_util::FutureExt;
 use reth::providers::BlockReaderIdExt;
-use reth::{providers::StateProviderFactory, tasks::TaskSpawner};
+use reth::providers::StateProviderFactory;
 use reth_basic_payload_builder::HeaderForPayload;
 use reth_basic_payload_builder::{BasicPayloadJobGeneratorConfig, PayloadConfig};
 use reth_node_api::PayloadBuilderAttributes;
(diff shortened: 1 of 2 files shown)


crates/op-rbuilder/src/generator.rs
```diff
@@ -260,24 +251,19 @@ where
 
         let (tx, rx) = oneshot::channel();
         self.build_complete = Some(rx);
+        let args = BuildArguments {
```
ferranbt, line 254: I imagined that it would be PayloadGenerator the one spawning the job instead of the PayloadBuilder impl.

SozinM: I could move this too into the OpPayloadBuilderVanilla. 
I will add config and cancel to the try_build and then see if we could refactor this somehow.

ferranbt: > I could move this too into the OpPayloadBuilderVanilla.

Forget about the highlighted line, I could not highlight the correct part of the code since it was not modified.

SozinM: Seems like i read initial message wrong.
You man to move spawning into new_payload_job of BlockPayloadJobGenerator, right?

SozinM: Ideally i would change interface in this place a bit.
I would like to see the following:
Instead of resolving per-job we would me payload resolving logic into `BlockPayloadJobGenerator`
The `BlockPayloadJobGenerator` would coordinate all `BlockPayloadJob` spawned for this payload building and will ultimately return response to `PayloadBuilderService`

SozinM:
Will do it differently on top of flashbots/rbuilder#443 chain

PR #442, opened 2025-02-21
Make flashblock ws url as a flag and add more data

cody-wang-cb:
## 📝 Summary
Making flashblock websocket as flag so that it can be customizable.

Also added receipts, account balance, block number into the metadata field, such that the RPC node can build the RPC response for `eth_transactionReceipt` and `eth_getBalance` for pending blocks. (https://github.com/danyalprout/reth-flashblocks/pull/4)

Added an integration test to test that the metadata field indeed has the new data.

* [x] Run `make lint`
* [x] Run `make test`


ZanCorDX:
Can you fix the checks?

crates/op-rbuilder/src/integration/integration_test.rs
```diff
@@ -83,12 +105,46 @@ mod tests {
                     .expect("receipt");
             }
         }
-
         // there must be a line logging the monitoring transaction
         op_rbuilder
-            .find_log_line("Committed block built by builder")
+            .find_log_line("Committed block built by builder") // no builder tx for flashblocks builder
```
cody-wang-cb: Looks like the flashblock builder does not include this additional builder tx as the vanilla builder. https://github.com/flashbots/rbuilder/blob/develop/crates/op-rbuilder/src/payload_builder_vanilla.rs#L508

Is this expected? Curious what's the reason that there is always a builder tx for the vanilla builder

crates/op-rbuilder/src/integration/integration_test.rs
```diff
@@ -198,7 +254,7 @@ mod tests {
                     .transactions
                     .hashes()
                     .any(|hash| hash == *reverting_tx.tx_hash()),
-                "reverted transaction unexpectedly included in block"
+                "reverted transaction unexpectedly included in block" // flashblock builder still includes reverted txs
```
cody-wang-cb: The flashblock builder also always include the tx with revert input. Seems like there might be some discrepancy for `execute_best_transactions` between the flashblock builder and the vanilla builder. But quite unclear to me yet by looking at the code. 
https://github.com/flashbots/rbuilder/blob/develop/crates/op-rbuilder/src/payload_builder.rs#L1015
https://github.com/flashbots/rbuilder/blob/develop/crates/op-rbuilder/src/payload_builder_vanilla.rs#L1140

Wondering if anyone can point me to the exact logic that prevent this revert tx to be included in the vanilla builder?

cody-wang-cb (commented):
@ZanCorDX @ferranbt got some questions while working on the tests, mind taking a look and provide some guidance?

.github/workflows/checks.yaml
```diff
@@ -171,13 +171,13 @@ jobs:
...
       - name: Build the rbuilder
-        run: cargo build -p op-rbuilder --bin op-rbuilder --features optimism
+        run: cargo build -p op-rbuilder --bin op-rbuilder --features optimism,flashblocks
 
       - name: Generate test genesis file
         run: cargo run -p op-rbuilder --bin tester --features optimism -- genesis --output genesis.json
 
       - name: Run integration tests
-        run: cargo test --package op-rbuilder --lib --features optimism,integration -- integration::integration_test::tests
+        run: cargo test --package op-rbuilder --lib --features optimism,integration,flashblocks -- integration::integration_test::tests
```
avalonche, line 186: could you also add the features to the makefile?

crates/op-rbuilder/src/payload_builder_vanilla.rs
```diff
@@ -76,15 +76,20 @@ use std::{fmt::Display, sync::Arc, time::Instant};
 use tokio_util::sync::CancellationToken;
 use tracing::{info, trace, warn};
 
-#[derive(Debug, Clone, Copy, Default)]
+#[derive(Debug, Clone, Default)]
 #[non_exhaustive]
 pub struct CustomOpPayloadBuilder {
     builder_signer: Option<Signer>,
+    #[allow(dead_code)]
```
avalonche: I think you can have this field enabled only if the flashblocks feature is enabled?

crates/op-rbuilder/src/args.rs
```diff
@@ -17,4 +17,11 @@ pub struct OpRbuilderArgs {
     /// Builder secret key for signing last transaction in block
     #[arg(long = "rollup.builder-secret-key", env = "BUILDER_SECRET_KEY")]
     pub builder_signer: Option<Signer>,
+    /// Websocket port for flashblock payload builder
+    #[arg(
+        long = "rollup.flashblocks-ws-url",
```
avalonche, line 22: wondering if it makes sense to separate the host and port like the rpc server?

crates/op-rbuilder/src/integration/integration_test.rs
avalonche: the builder uses the builder tx for monitoring purposes, see [here](https://github.com/flashbots/rbuilder/blob/develop/crates/op-rbuilder/README.md#observability). These integration tests are for the vanilla builder, not flashblocks. 
I think it would make more sense to have a different set of integration tests for flashblocks. I think you can configure this in rust like
```
  // This test always runs
  #[test]
  fn always_runs() {
      // Test code here
  }
  
  // This test only runs when "my_feature" is enabled
  #[cfg(feature = "my_feature")]
  #[test]
  fn only_with_feature() {
      // Test code that depends on my_feature
  }
```

crates/op-rbuilder/src/integration/integration_test.rs
avalonche: https://github.com/flashbots/rbuilder/blob/develop/crates/op-rbuilder/src/payload_builder_vanilla.rs#L1223
vanilla builder skips reverting transactions and removes them from the tx pool. I recommend having a separate set of integration tests for the flashblocks builder

.github/workflows/checks.yaml
avalonche: actually if you want the tests to pass you should remove flashblocks from the integration tests

crates/op-rbuilder/src/integration/integration_test.rs
cody-wang-cb: Yeah it makes sense that these tests were built for the vanilla builder. I could definitely move these new tests logics to a new test under `flashblocks` feature.
But I wonder for op-rbuilder, shouldn't the flashblock builder be the default builder? It seems a bit weird to me that there's discrepancy for the two builders in terms of functionality, besides the subblock part

crates/op-rbuilder/src/payload_builder.rs
```diff
@@ -548,10 +554,30 @@ where
...
+        .zip(new_receipts.iter())
+        .map(|(tx, receipt)| (*tx.tx_hash(), receipt.clone()))
+        .collect::<HashMap<B256, N::Receipt>>();
+    let new_account_balances = new_bundle
+        .state
+        .iter()
+        .filter_map(|(address, account)| account.info.as_ref().map(|info| (*address, info.balance)))
+        .collect::<HashMap<Address, U256>>();
+
+    let metadata = serde_json::json!({
```
avalonche: can add a new metadata struct with serde encoding

.github/workflows/checks.yaml
cody-wang-cb: I think I should add another CI check for the flashblocks builder tests? (added above)

crates/op-rbuilder/src/args.rs
cody-wang-cb: I think it's ok for now, if people want to change the port they can just pass a diff url

crates/op-rbuilder/src/integration/integration_test.rs
avalonche: flashblocks is still not ready for production so we are running the vanilla block builder as the default for now. I think eventually the plan is to make it default

.github/workflows/checks.yaml
avalonche: if its guarded behind a feature flag you should need a separate command / CI test and everything should pass

cody-wang-cb: yup the above is a separate CI test in addition to the original test

avalonche (approved):

PR #443, opened 2025-02-24
Move out payload builder service from reth

SozinM:
## 📝 Summary

This PR reduces our reliance on Reth by taking direct ownership of the main block-building component. Doing so prevents potential disruptions from future Reth interface changes, ensuring a more stable and predictable development process.

## 💡 Motivation and Context

---

* [ ] Run `make lint`
* [ ] Run `make test`


crates/op-rbuilder/src/primitives/payload_builder_service.rs
avalonche, line 1: what's the plan to maintain upstream changes with Reth's payload building service?

SozinM: i think with this approach it's possible to us to fix this struct and stop getting upstream changes
It will work until fundamental changes, that change model of engine_FCU and engine_getPayload

PR #444, opened 2025-02-24
Move ExecutionInfo and ExecutedPayload into different crate

SozinM:
## 📝 Summary

Moving out types that used in both builders to simplify it.
Goes after https://github.com/flashbots/rbuilder/pull/443

Clear PR with changes: https://github.com/NethermindEth/rbuilder/pull/5
## 💡 Motivation and Context

---

* [ ] Run `make lint`
* [ ] Run `make test`

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -6724,6 +6724,7 @@ dependencies = [
  "tikv-jemallocator",
  "time",
  "tokio",
+ "tokio-stream",
  "tokio-tungstenite 0.26.1",
  "tokio-util",
  "tower 0.4.13",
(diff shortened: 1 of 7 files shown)


avalonche (commented):
this seems to contain changes from https://github.com/flashbots/rbuilder/pull/443?

avalonche (approved):

avalonche:
this looks fine to me, just wondering about maintenance - since we are moving reth types out of the file, what would the process be like for comparing the diff between rbuilder and reth to ensure reth changes are upstreamed correctly?

SozinM:
@avalonche Good question.
Problems i see in the current code:
1. We have type duplicates
2. Some types have different order from reth builder.rs

How i see process of rebasing:
We open diff between reth , for example 1.1.5 <> 1.2.0
We go trough all changes in payload/src/builder.rs
And if we see changes we need to take we put them in relevant files

ferranbt (approved):

PR #457, opened 2025-03-03
Performance micro-optimizations

mralj:
## 📝 Summary

The PR contains small perf micro-optimisations. 

### Dashmap instead  of Mutex<Hashmap>
Given concurrent access to `Noncer`, and the fact that per block there is non-trivial amount of addresses being cached, `Dashmap` should offer better perf than `Mutex<Hashmap>`. 

Additionally, IMO, the Mutex lock on Noncer was held _too long_: 

```Rust
   let mut cache = self.cache.lock();
        if let Some(nonce) = cache.get(&address) {
            return Ok(*nonce);
        }
         // THIS HERE IS POTENTIALLY PROBLEMATIC
        // If this is disk read (in case account isn't cached) or, when remote state provider is implemented, this can take time
        // depending on disk speed/remote state provider we are looking at up to a couple hundred micro_sec  
        let nonce = self.state.account_nonce(&address)?.unwrap_or_default();
        cache.insert(address, nonce);
        Ok(nonce)
```

There is one additional micro-optimisation possible here, and that is usage of something like _no-hash hasher_. Since ETH Addresses are already derived by hashing, we don't have to rehash them, but this seemed like overkill :) 

### parking_lot instead of sync::Mutex
The `rbuilder` codebase generally uses `parking_lot` Mutex. AFAIK it should have better perf than `sync` one, and also it doesn't have poisoning (so no `unwrap()` is necessary)

---

* [x] Run `make lint`
* [x] Run `make test`


dvush (approved):
Thanks! I like the dash map change, I agree that lock was non-optimal. 
Parking lot is also nice.

PR #459, opened 2025-03-04
Fix missing slot regression, better track of payload attributes in logs

dvush:
* It introduces unique `payload_id` for each payload received from the CL
* Multiple logs are added to track possible reasons for not building with the given `payload_id`

It was found that in current develop we miss approx. 5% of the slots because we don't cancel old jobs mid slot anymore when reth goes to the next block. 

This adds a new job for each slot `run_check_if_parent_block_is_last_block` that explicitly checks last reth block every 100ms and cancels payload if we can't use it anymore. This fixes the issue above and makes us more robust since before we relied on reth giving us specific error type to cancel slot when reth updated last block and now we check it explicitly.

---

* [ ] Run `make lint`
* [ ] Run `make test`


ZanCorDX (approved):
Can payload_id be incremental instead of rand::random()?

PR #462, opened 2025-03-05
Interop support

SozinM:
Cherry-pick kona structs that we are using for interop validation
Implement SupervisorValidator to be used in rbuilder 
Add additional primitives crate, that is used for storing external code

## 💡 Motivation and Context

---

* [x] Run `make lint`
* [x] Run `make test`

### DIFF
--- a/crates/op-rbuilder/src/payload_builder.rs
+++ b/crates/op-rbuilder/src/payload_builder.rs
@@ -75,6 +75,7 @@ use tokio::net::{TcpListener, TcpStream};
 use tokio::sync::mpsc;
 use tokio_tungstenite::accept_async;
 use tokio_tungstenite::WebSocketStream;
+use url::Url;
 
 use serde::{Deserialize, Serialize};
 
@@ -88,11 +89,15 @@ struct FlashblocksMetadata<N: NodePrimitives> {
 #[derive(Debug, Clone, Default)]
 #[non_exhaustive]
 pub struct CustomOpPayloadBuilder {
-    #[allow(dead_code)]
+    #[expect(dead_code)]
     builder_signer: Option<Signer>,
     flashblocks_ws_url: String,
     chain_block_time: u64,
     flashblock_block_time: u64,
+    #[expect(dead_code)]
+    supervisor_url: Option<Url>,
+    #[expect(dead_code)]
+    supervisor_safety_level: Option<String>,
 }
 
 impl CustomOpPayloadBuilder {
@@ -101,12 +106,16 @@ impl CustomOpPayloadBuilder {
         flashblocks_ws_url: String,
         chain_block_time: u64,
         flashblock_block_time: u64,
+        supervisor_url: Option<Url>,
+        supervisor_safety_level: Option<String>,
     ) -> Self {
         Self {
             builder_signer,
             flashblocks_ws_url,
             chain_block_time,
             flashblock_block_time,
+            supervisor_url,
+            supervisor_safety_level,
         }
     }
 }
--- a/crates/op-rbuilder/src/payload_builder_vanilla.rs
+++ b/crates/op-rbuilder/src/payload_builder_vanilla.rs
@@ -1,9 +1,8 @@
-use crate::generator::BlockPayloadJobGenerator;
-use crate::generator::BuildArguments;
 use crate::{
-    generator::{BlockCell, PayloadBuilder},
+    generator::{BlockCell, BlockPayloadJobGenerator, BuildArguments, PayloadBuilder},
     metrics::OpRBuilderMetrics,
     primitives::reth::{ExecutedPayload, ExecutionInfo},
+    primitives::supervisor::SupervisorValidator,
     tx_signer::Signer,
 };
 use alloy_consensus::constants::EMPTY_WITHDRAWALS;
@@ -13,9 +12,12 @@ use alloy_consensus::{
 };
 use alloy_eips::merge::BEACON_NONCE;
 use alloy_primitives::private::alloy_rlp::Encodable;
-use alloy_primitives::{Address, Bytes, TxHash, TxKind, U256};
+use alloy_primitives::{Address, Bytes, TxHash, TxKind, B256, U256};
 use alloy_rpc_types_engine::PayloadId;
 use alloy_rpc_types_eth::Withdrawals;
+use jsonrpsee::http_client::HttpClientBuilder;
+use kona_interop::{ExecutingDescriptor, SafetyLevel};
+use kona_rpc::{InteropTxValidator, InteropTxValidatorError};
 use op_alloy_consensus::{OpDepositReceipt, OpTypedTransaction};
 use reth::builder::{components::PayloadServiceBuilder, node::FullNodeTypes, BuilderContext};
 use reth::core::primitives::InMemorySize;
@@ -76,7 +78,8 @@ use revm::{
 use std::error::Error as StdError;
 use std::{fmt::Display, sync::Arc, time::Instant};
 use tokio_util::sync::CancellationToken;
-use tracing::{info, trace, warn};
+use tracing::{error, info, trace, warn};
+use url::Url;
 
 #[derive(Debug, Clone, Default)]
 #[non_exhaustive]
@@ -88,6 +91,8 @@ pub struct CustomOpPayloadBuilder {
     chain_block_time: u64,
     #[cfg(feature = "flashblocks")]
     flashblock_block_time: u64,
+    supervisor_url: Option<Url>,
+    supervisor_safety_level: Option<String>,
 }
 
 impl CustomOpPayloadBuilder {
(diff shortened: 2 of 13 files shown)


SozinM:
Don't look at the failing lint in this PR, it will be fixed with https://github.com/flashbots/rbuilder/pull/465

SozinM:
https://discord.com/channels/1244729134312198194/1332034583394254929/1346955305904373830
Blocked by this

crates/op-rbuilder/src/payload_builder_vanilla.rs
```diff
@@ -1104,6 +1134,30 @@ where
...
+                            self.metrics.inc_num_cross_chain_tx_timeout();
+                            continue;
+                        }
+                        err => {
+                            trace!(target: "payload_builder", %err, ?sequencer_tx, "Executing message rejected.");
+                            self.metrics.inc_num_cross_chain_tx_fail();
+                            continue;
+                        }
+                    },
+                }
```
emhane: there is also the client error which you probably want to match on specifically, because depending on what it is, you may want to try and reconnect with the supervisor

crates/op-rbuilder/src/primitives/kona/access_list.rs
```diff
@@ -0,0 +1,27 @@
+//! [Source](https://github.com/op-rs/kona/blob/0d0d390ea989c0ac26ae67a522903b958c1cf011/crates/protocol/interop/src/access_list.rs)
```
emhane: copyright missing

crates/op-rbuilder/src/primitives/kona/api.rs
```diff
@@ -0,0 +1,49 @@
+//! [Source](https://github.com/op-rs/kona/blob/a1d8ea603960cb4bd3cc19784f7c3365352f1849/crates/node/rpc/src/api.rs)
```
emhane: copyright license missing

crates/op-rbuilder/src/primitives/kona/error.rs
```diff
@@ -0,0 +1,143 @@
+//! Interop supervisor RPC client errors.
```
emhane: copyright license missing

crates/op-rbuilder/src/primitives/kona/message.rs
```diff
@@ -0,0 +1,20 @@
+//! [Source](https://github.com/op-rs/kona/blob/a1d8ea603960cb4bd3cc19784f7c3365352f1849/crates/protocol/interop/src/message.rs)
```
emhane: copyright license missing

crates/op-rbuilder/src/primitives/kona/safety.rs
```diff
@@ -0,0 +1,41 @@
+//! [Source](https://github.com/op-rs/kona/blob/a1d8ea603960cb4bd3cc19784f7c3365352f1849/crates/protocol/interop/src/safety.rs)
```
emhane: copyright license missing

crates/op-rbuilder/src/primitives/kona/traits.rs
```diff
@@ -0,0 +1,52 @@
+//! RPC validator component used in interop.
```
emhane: copyright license missing

emhane (changes requested):
you're missing copyright license on a lot of files, read from copyright licence in kona:
> The above copyright notice and this permission notice shall be included in all copies or substantial portions of the Software.

https://github.com/op-rs/kona/blob/main/LICENSE.md

SozinM:
@emhane 
Thanks! Right now i update rust version and could use kona itself and not copy the code, working on it now

crates/op-rbuilder/src/primitives/kona/access_list.rs
SozinM: Fixed all of those

crates/op-rbuilder/src/payload_builder.rs
```diff
@@ -93,6 +94,8 @@ pub struct CustomOpPayloadBuilder {
     flashblocks_ws_url: String,
     chain_block_time: u64,
     flashblock_block_time: u64,
+    #[allow(dead_code)]
```
emhane:
```suggestion
    #[expect(dead_code)]
```
prefer using `expect` so compiler helps you to clean up redundant attributes

crates/op-rbuilder/src/payload_builder.rs
```diff
@@ -101,12 +104,14 @@ impl CustomOpPayloadBuilder {
         flashblocks_ws_url: String,
         chain_block_time: u64,
         flashblock_block_time: u64,
+        supervisor_url: Option<Url>,
```
emhane, line 109: the right way to do builder pattern is not to include this optional attribute in constructor, instead have it as a separate method
```rust
pub fn supervisor_url(mut self, url: Url) -> Self {
    self.supervisor_url = Some(url);
    self
}
```

crates/op-rbuilder/src/payload_builder_vanilla.rs
```diff
@@ -143,6 +155,7 @@ where
             pool,
             ctx.provider().clone(),
             Arc::new(BasicOpReceiptBuilder::default()),
+            self.supervisor_url.clone(),
```
emhane, line 161:
```suggestion
            Some(ctx.provider()
                .chain_spec()
                .is_interop_active_at_timestamp(ctx.provider().chain_spec().genesis_header().timestamp())
                .then(|| self.supervisor_url().ok_or(PayloadBuilderError::SupervisorUrlMissingPostInterop)?),
```
smthg like this

crates/op-rbuilder/src/payload_builder_vanilla.rs
```diff
@@ -1355,3 +1410,88 @@ fn estimate_gas_for_builder_tx(input: Vec<u8>) -> u64 {
...
+
+/// Extracts commitment from access list entries, pointing to 0x420..022 and validates them
+/// against supervisor.
+///
+/// If commitment present pre-interop tx rejected.
+///
+/// Returns (is_valid, is_recoverable)
+pub fn is_cross_tx_valid<N: NodePrimitives>(
+    tx: &N::SignedTx,
+    client: Option<&SupervisorValidator>,
```
emhane, line 1449:
```suggestion
    client: &SupervisorValidator,
```
it seems like the caller is responsible if checking if interop is activated, so supervisor client isn't an optional param

SozinM: I want to simplify execute_best_transactions as much as possible, it would look consistent 
Don't the same for reth PR

SozinM: In the reth PR I will ensure that the supervisor is set during the building, but in here we need to rebase first

emhane: makes sense

crates/op-rbuilder/src/payload_builder_vanilla.rs
```diff
@@ -1355,3 +1410,88 @@ fn estimate_gas_for_builder_tx(input: Vec<u8>) -> u64 {
 
     zero_cost + nonzero_cost + 21_000
 }
+
+/// Extracts commitment from access list entries, pointing to 0x420..022 and validates them
+/// against supervisor.
+///
+/// If commitment present pre-interop tx rejected.
+///
+/// Returns (is_valid, is_recoverable)
```
tynes, line 1445: What exactly is `is_recoverable`? Is it like an RPC error?

SozinM: we remove totally invalid transactions from mempool
And if the transaction is not yet valid - just skip it for the current round

tynes:
I see that the supervisor client calls `check_messages` - it should be calling `supervisor_checkAccessList` per https://github.com/ethereum-optimism/specs/pull/612

We probably want to enable a configurable safety level that is used in the client. The safety level could in theory be configured on a per chain or per message basis (accept any unsafe message if value is less than $100 for example) but that is def over engineering for now. I would be ok with either hardcoding the safety level to unsafe for now or allowing a runtime flag that allows it to be set to one of the possible values

SozinM:
@tynes thanks for review a lot!
We are using [kona](https://github.com/op-rs/kona/blob/f7b2cf21588140fb8ea98aed72cce65a19fcc13e/crates/node/rpc/src/interop/mod.rs#L49) to validate messages and it's using `supervisor_checkAccessList`.
As for safety levels - before we were using cross-unsafe, and my experiences with unsafe were that any message (even non-existing) passes as unsafe, unless it grossly malformed, have it changed? 
It's definitely a good idea to make it configurable, thank you!

ferranbt (approved):

PR #469, opened 2025-03-06
fix optional bundles txs, add metric for first seen at, count unique bundles modulo replacements

dvush:
## 📝 Summary

see title

* [ ] Run `make lint`
* [ ] Run `make test`

### DIFF
--- a/crates/rbuilder/src/primitives/serialize.rs
+++ b/crates/rbuilder/src/primitives/serialize.rs
@@ -137,7 +137,7 @@ pub struct RawBundle {
     /// calculated. Defaults to final transaction in the bundle if list is not specified/empty.
     #[serde(skip_serializing_if = "Option::is_none")]
     pub refund_tx_hashes: Option<Vec<TxHash>>,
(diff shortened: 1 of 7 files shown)


crates/rbuilder/src/primitives/serialize.rs
```diff
@@ -137,7 +137,7 @@ pub struct RawBundle {
     /// calculated. Defaults to final transaction in the bundle if list is not specified/empty.
     #[serde(skip_serializing_if = "Option::is_none")]
     pub refund_tx_hashes: Option<Vec<TxHash>>,
-    /// firstSeenAt `Number`, timestamp at which bundle was first seen,
+    /// firstSeenAt `Number`, timestamp (in milliseconds) at which bundle was first seen,
```
ZanCorDX: Isn't the timestamp in seconds?

ZanCorDX (approved):

crates/rbuilder/src/primitives/serialize.rs
dvush: fixed!

PR #470, opened 2025-03-06
Fix payload parent header fetch

dvush:
## 📝 Summary

Fixes a bug where we were not even trying to fetch last block header
from reth database if request arrived after the fetch deadline. It was the reason for
missing 0.2% of slots.

Now we always fetch at least once.

---

* [ ] Run `make lint`
* [ ] Run `make test`


ryanschneider (approved):
lgtm!

PR #482, opened 2025-03-10
Make block times dynamic in flashblock builder

cody-wang-cb:
## 📝 Summary
Currently the block time is hard coded to assume the chain block time is 1s and flashblock time is 250ms. But for Base the chain block time is 2s. Thus fixing this by adding args to make it dynamic.

---

* [x] Run `make lint`
* [x] Run `make test`

### DIFF
--- a/crates/op-rbuilder/src/args.rs
+++ b/crates/op-rbuilder/src/args.rs
@@ -25,9 +25,17 @@ pub struct OpRbuilderArgs {
     )]
     pub flashblocks_ws_url: String,
     /// chain block time in milliseconds
-    #[arg(long = "rollup.chain-block-time", default_value = "1000")]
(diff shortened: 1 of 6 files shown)


avalonche (approved):

crates/op-rbuilder/src/args.rs
```diff
@@ -24,4 +24,10 @@ pub struct OpRbuilderArgs {
         default_value = "127.0.0.1:1111"
     )]
     pub flashblocks_ws_url: String,
+    /// chain block time in milliseconds
+    #[arg(long = "rollup.chain-block-time", default_value = "1000")]
```
avalonche: can add env here to make this configurable via env variables

PR #485, opened 2025-03-11
Download sccache manually in Dockerfile

SozinM:
It's not recommended to cargo install sccache per mozilla:
```
Note that this will compile sccache from source which is fairly resource-intensive. For CI purposes you should use prebuilt binary packages.
```
Added manual download, supporting platforms linux/amd64 and linux/arm64

## 💡 Motivation and Context

---

* [x] Run `make lint`
* [x] Run `make test`


sukoneck (approved):
Thank you!

PR #489, opened 2025-03-11
IPC state provider

mralj:
## 📝 Summary

`rbuilder` can now work with _any node_ via IPC. 
This means that any node can provide the state for `rbuilder`,  `revm` is still used as the EVM. 
Node requirements are that it exposes the following RPC calls: 

Calls not available in the [ETH JSON RPC spec:](https://ethereum.github.io/execution-apis/api-documentation/) 
1. `rbuilder_calculateStateRoot` for state root calculation
2. `rbuilder_getCodeByHash` given Bytecode hash, returns Bytecode

Calls optimised for `rbuilder`, but have a counterpart in the ETH  JSON RPC spec: 

3. `rbuilder_getBlockHash` gets the block hash  given block number (similar to `eth_getBlockByNumber`, but just returns block hash)
4. `rbuilder_getAccount` gets account info (similar to `eth_getProof`, but w/o _proof stuff_)

To use `rbuilder` with node via IPC,  the `config.toml` must have the following config (example):

```TOML
[ipc_provider]
ipc_path = "/root/execution-data/nethermind.ipc"
mempool_server_url = "ws://localhost:8546"
request_timeout_ms = 75
```

### IPC

This implementation was initially intended to introduce a _remote state provider_. By remote, I mean that the idea was that state could be provided via HTTP/WS/IPC. Unfortunately, due to implementation issues/constraints, I've decided only to implement state provisioning via IPC.
I don't think this has any practical downside, especially since the state provider must be fast. There is a non-trivial number of calls to read state (~300/s), meaning it would be unrealistic to use this over the network and have near disk read latency. 

Code-wise, issues above and constraints stem mainly from the fact that the traits to read state are `sync`. Initially, I relied on tokio and alloy.rs to fetch this remote state, but this implementation had many issues.  
Firstly, each call to fetch any data (e.g. fetching account or Bytecode) had to be wrapped in a function call like this:

```Rust
    /// Runs fututre in sync context
    // StateProvider(Factory) traits require sync context, but calls to remote provider are async
    // What's more, rbuilder is executed in async context, so we have a situation
    // async -> sync -> async
    // This helper function allows execution in such environment
    fn run<F, R>(&self, f: F) -> R
    where
        F: Future<Output = R>,
    {
        tokio::task::block_in_place(|| self.runtime_handle.block_on(f))
    }
```

This adds additional overhead on Tokio runtime and doesn't play well with some parts of the codebase, specifically mutex locking. Not to go too deep into explaining issues around this in the PR description, but we would end up in scenarios where the whole Tokio runtime I/O would be blocked or we would dead-lock parking-lot mutexes (in some scenarios).

The solutions (_[monitoring thread](https://github.com/tokio-rs/tokio/issues/4730)_ + async mutex) seemed hacky and suboptimal. 
This is why, in the end, I reached for sync (but concurrent) IPC solution, which I implemented from scratch [here. It's called REIPC (coming from request/response IPC)](https://github.com/NethermindEth/reipc). 

This solution was tested using Nethermind node, and while [Nethermind will do some improvements](https://github.com/NethermindEth/nethermind/issues/8258#issue-2875081273) on IPC to reduce latency and increase throughput, here are the initial request&response latencies: 

CleanShot 2025-03-01 at 11 44 11@2x

### Dashmap & QuickCache in IPC provider
We need caches because otherwise the number of IPC calls would be pretty high (in thousands per sec). 
I also reached for concurrent caches because both `StateProviderFactory` and `StateProvider` need to be `Send + Sync`.

QuickCache is used so that I don't have to implement concurrent-cache invalidation by hand :) 
[Here](https://github.com/moka-rs/moka?tab=readme-ov-file#choosing-the-right-cache-for-your-use-case) is some info on QuickCache vs Moka (other popular caching crate), TL;DR; for our simple case QuickCache seems a better fit.

### On `enum StateProviderFactories`
~The reason `cli.rs` passes `StateProviderFactories` enum to `config.new_builder` is because I wanted the state provider for `rbuilder` to be chosen via config at the runtime.~
~This is why, AFAIK, static dispatch is not an option.  So I was left with the choice of refactoring to dynamic dispatch (akin to `StateProviderBox`) OR the `enum` solution.~
~I chose the `enum` solution for following reasons:~
~1. It seemed to me that code diff would be smaller (smaller change to implement)~
~2. AFIAK it's faster. Enum matching will compile to `JMP` vs `CALL` (in case of dynamic dispatch), which compiler will be able to optimize better (especially in this scenario), and given branch predicitoning I guess that it'll be almost free (+ no need for vtable/pointer loading).]~

UPDATE: this was properly handled [in this commit](https://github.com/flashbots/rbuilder/pull/489/commits/2dc4f41e7f77e605502ce9da1148d8b912bd0de0) thanks to suggestions from @ZanCorDX 

### On `MempoolSource` enum

The reason I chose to use WebSockets for streaming transactions when `rbuilder` uses IPC state provider is because, currently, REIPC doesn't support ETH subscriptions/streams.

* [x] Run `make lint`
* [x] Run `make test`

### DIFF
--- a/crates/rbuilder/src/provider/ipc_state_provider.rs
+++ b/crates/rbuilder/src/provider/ipc_state_provider.rs
@@ -66,7 +66,7 @@ impl IpcStateProviderFactory {
 #[derive(Debug, Clone, Deserialize, PartialEq, Eq)]
 #[serde(default, deny_unknown_fields)]
 pub struct IpcProviderConfig {
-    pub(crate) request_timeout: Duration,
+    pub(crate) request_timeout: u64,
     pub(crate) ipc_path: PathBuf,
     pub(crate) txpool_server_url: String,
 }
@@ -74,7 +74,7 @@ pub struct IpcProviderConfig {
 impl Default for IpcProviderConfig {
     fn default() -> Self {
         Self {
-            request_timeout: Duration::from_millis(DEFAULT_IPC_REQUEST_TIMEOUT_MS),
+            request_timeout: DEFAULT_IPC_REQUEST_TIMEOUT_MS,
             txpool_server_url: String::new(),
             ipc_path: PathBuf::new(),
         }
--- a/crates/rbuilder/src/provider/ipc_state_provider.rs
+++ b/crates/rbuilder/src/provider/ipc_state_provider.rs
@@ -1,4 +1,9 @@
-use std::{fmt::Debug, path::Path, sync::Arc, time::Duration};
+use std::{
+    fmt::Debug,
+    path::{Path, PathBuf},
+    sync::Arc,
+    time::Duration,
+};
 
 use alloy_consensus::{constants::KECCAK_EMPTY, Header};
 use alloy_eips::{BlockId, BlockNumHash, BlockNumberOrTag};
@@ -30,7 +35,7 @@ use super::{RootHasher, StateProviderFactory};
 /// 100ms was picked up after initial testing using Nethermind client as state provider
 /// 99.9% requests return within 50ms; using 100ms gives us error rate of ~0.03%
 /// Median response time is ~300 micro_sec.
-const IPC_REQUEST_TIMEOUT_MS: u64 = 100;
+const DEFAULT_IPC_REQUEST_TIMEOUT_MS: u64 = 100;
 
 /// Remote state provider factory allows providing state via remote RPC calls over IPC
 /// Specifically UnixDomainSockets
@@ -44,19 +49,34 @@ pub struct IpcStateProviderFactory {
 }
 
 impl IpcStateProviderFactory {
-    pub fn new(ipc_path: &Path) -> Self {
-        let ipc_provider = RpcProvider::try_connect(
-            ipc_path,
-            Duration::from_millis(IPC_REQUEST_TIMEOUT_MS).into(),
-        )
-        // there is no need to gracefully handle (or propagate) this error, if we cannot connect
-        // to IPC, then rbuilder cannot work
-        .expect("Can't connect to IPC");
+    pub fn new(ipc_path: &Path, req_timeout: Duration) -> Self {
+        let ipc_provider = RpcProvider::try_connect(ipc_path, req_timeout.into())
+            // there is no need to gracefully handle (or propagate) this error, if we cannot connect
+            // to IPC, then rbuilder cannot work
+            .expect("Can't connect to IPC");
 
         Self {
             ipc_provider,
             code_cache: Arc::new(DashMap::new()),
             state_provider_by_hash: Arc::new(DashMap::new()),
+        }
+    }
+}
+
+#[derive(Debug, Clone, Deserialize, PartialEq, Eq)]
+#[serde(default, deny_unknown_fields)]
+pub struct IpcProviderConfig {
+    pub(crate) request_timeout: Duration,
+    pub(crate) ipc_path: PathBuf,
+    pub(crate) txpool_server_url: String,
+}
+
+impl Default for IpcProviderConfig {
+    fn default() -> Self {
+        Self {
+            request_timeout: Duration::from_millis(DEFAULT_IPC_REQUEST_TIMEOUT_MS),
+            txpool_server_url: String::new(),
+            ipc_path: PathBuf::new(),
         }
     }
 }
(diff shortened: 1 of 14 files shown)


crates/rbuilder/src/provider/ipc_state_provider.rs
```diff
@@ -0,0 +1,534 @@
...
+
+    /// Gets block number of latest known block
+    fn best_block_number(&self) -> ProviderResult<BlockNumber> {
+        self.last_block_number()
+    }
+
+    /// Gets block header given block hash
+    fn header_by_number(&self, num: u64) -> ProviderResult<Option<Header>> {
+        let block = rpc_call::<
+            _,
```
SozinM: I'm in favor of using concrete type here. This will help in case we will do any refactoring and accidentally change parameters. This way we would be able to find problem instantly.

dvush:
Great! Thanks, I'll review it this week. (cc @ZanCorDX  don't merge before pls)

mralj:
Btw @dvush as stated in "On enum StateProviderFactories" I wanted for `rbuilder` to be completely configurable via `Config.toml`, but if you guys don't like this, I'm perfectly fine removing this, and handling this via feature flags  on comptime :)

crates/rbuilder/src/provider/ipc_state_provider.rs
mralj: Sorry, I completely missed the comment 🤦‍♂️
I disagree; the type here is inferred from the function arguments, i.e., `(num, false)`. ' If we mess it up, rust will yell at us. 
That being said, I don't feel strongly about this if @ZanCorDX  agrees with you, I'll refactor :)

SozinM: hm, will it yell if we change params to (num, false, false) for example? If yes - all good
But i assume it will send them to rpc and get error because there is extra false param

mralj: Correct, it won't :)
So the idea is that if we have written this out in 2 places (in the type and in the args) there is smaller chance of us making mistake in both places when refactoring?

ie. If we change one and not the other compiler will give us error, and we'll stop to think about i (hopefully not just copying args to type, because then compiler will tell you everything is ok and you'll still get rpc error xD) 

But all in all I agree this reduces the chance of developer error, I'll fix it :)

crates/rbuilder/src/provider/ipc_state_provider.rs
```diff
@@ -0,0 +1,534 @@
...
+const DEFAULT_STATE_CACHE_SIZE: usize = 128;
+
+/// Remote state provider factory allows providing state via remote RPC calls over IPC
+/// Specifically UnixDomainSockets
+#[derive(Clone, Debug)]
+pub struct IpcStateProviderFactory {
+    ipc_provider: RpcProvider,
+
+    code_cache: Arc<DashMap<B256, Bytecode>>,
+    state_provider_by_hash: Arc<Cache<BlockHash, Arc<IpcStateProvider>>>,
```
dvush, line 58: Can you please clarify for me the difference between DashMap and Cache here?

dvush (approved):
Awesome, pretty clear and isolated change. I've been running it locally and it works fine.

The only thing that I notices is root hash is much slower on this version. I can give more data tomorrow when I run the default rbuilder. But if feels like its 2-3x slower.


I think our root hash implementation (eth-sparse-mpt) is quite good for the block building purposes and it will be much better than generic root hash when we finish all the possible improvements and it can be used for integrations with other clients through API. The only thing that it needs is ability to fetch proofs using path. Instead of default eth_getProof that asks for address we need to fetch proof by hash(address) and hash(slot) for the given hash(address)/address

mralj:
> I think our root hash implementation (eth-sparse-mpt) is quite good for the block building purposes and it will be much better than generic root hash when we finish all the possible improvements and it can be used for integrations with other clients through API. The only thing that it needs is ability to fetch proofs using path. Instead of default eth_getProof that asks for address we need to fetch proof by hash(address) and hash(slot) for the given hash(address)/address.

Thanks for pointing this out, I'll discuss with my (NMC) team and see how we can leverage this, I'll also add some logging to our node to see how much milli seconds State Root Calculation actually takes and will get back to you on numbers :)

crates/rbuilder/src/provider/ipc_state_provider.rs
mralj: `code_cache` can live as long as `rbuilder` lives, i.e., it doesn't have to be invalidated because it will stay the same no matter the current state.

On the other hand, we need to  clean `state_provider_by_hash` after we are _done_  with the StateProvider for _some block_, because after slot is finished, this data becomes useless and just grows. 
The reason I went with `Cache` here instead of `Dashmap` is because I didn't want to write concurrent cache invalidation myself :)

crates/rbuilder/src/provider/ipc_state_provider.rs
mralj: @SozinM done :)
Sorry it took me so long!

https://github.com/flashbots/rbuilder/pull/489/commits/fbe50877205181b7f961afd17379c043b6c339d1

PR #500, opened 2025-03-13
Add txn monitoring in op-rbuilder pool

ferranbt:
## 📝 Summary

This PR introduces a monitoring service that prints all the transactions that arrive to the transaction pool in op-rbuilder.

## 💡 Motivation and Context

---

* [ ] Run `make lint`
* [ ] Run `make test`

### DIFF
--- a/crates/op-rbuilder/src/args.rs
+++ b/crates/op-rbuilder/src/args.rs
@@ -40,6 +40,6 @@ pub struct OpRbuilderArgs {
     pub flashblock_block_time: u64,
 
     /// Signals whether to log pool transactions events
-    #[arg(long = "log-pool-transactions", default_value = "false")]
+    #[arg(long = "builder.log-pool-transactions", default_value = "false")]
     pub log_pool_transactions: bool,
 }
--- a/crates/op-rbuilder/src/args.rs
+++ b/crates/op-rbuilder/src/args.rs
@@ -38,4 +38,8 @@ pub struct OpRbuilderArgs {
         env = "FLASHBLOCK_BLOCK_TIME"
     )]
     pub flashblock_block_time: u64,
+
+    /// Signals whether to log pool transactions events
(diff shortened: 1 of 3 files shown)


crates/op-rbuilder/src/args.rs
```diff
@@ -38,4 +38,8 @@ pub struct OpRbuilderArgs {
         env = "FLASHBLOCK_BLOCK_TIME"
     )]
     pub flashblock_block_time: u64,
+
+    /// Signals whether to log pool transactions events
+    #[arg(long = "log-pool-transactions", default_value = "false")]
```
avalonche: non blocking: we should rename the args to the builder namespace

crates/op-rbuilder/src/monitor_tx_pool.rs
```diff
@@ -0,0 +1,43 @@
...
+pub async fn monitor_tx_pool(mut new_transactions: AllTransactionsEvents<OpPooledTransaction>) {
+    while let Some(event) = new_transactions.next().await {
+        transaction_event_log(event);
+    }
+}
+
+fn transaction_event_log(event: FullTransactionEvent<OpPooledTransaction>) {
+    match event {
+        FullTransactionEvent::Pending(hash) => {
+            debug!("Transaction event: tx={:?}, kind=pending", hash)
```
avalonche: this wouldn't work with json structured logging?
```suggestion
            debug!(
                    tx_hash=tx_hash.to_string(),
                    kind = "pending",
                    "Transaction event received"
                )
```

avalonche (approved):

crates/op-rbuilder/src/monitor_tx_pool.rs
ferranbt: I will change this once I am back at home

PR #507, opened 2025-03-16
deps: reth v1.3.4

efbig:
## 📝 Summary

Bump deps to reth `v1.3.3`

* [x] Run `make lint`
* [x] Run `make test`

### DIFF
--- a/crates/op-rbuilder/src/integration/integration_test.rs
+++ b/crates/op-rbuilder/src/integration/integration_test.rs
@@ -1,24 +1,24 @@
 #[cfg(all(test, feature = "integration"))]
 mod tests {
-    use crate::integration::{
-        op_rbuilder::OpRbuilderConfig, op_reth::OpRethConfig, IntegrationFramework,
+    use crate::{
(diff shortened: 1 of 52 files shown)


SozinM:
Looks good, i checked reth diff between 1.2.2 and 1.3.2 and didn't saw any logic changes (additional validations, etc) , so this should work.
I will do another look after rebase.
Great work!

crates/op-rbuilder/src/integration/integration_test.rs
```diff
@@ -314,7 +314,7 @@ mod tests {
 
         // Query the block and check transaction ordering
         let block = provider
-            .get_block_by_hash(block_hash, BlockTransactionsKind::Full)
```
SozinM, line 317: here we should add .full() after the function
```
    ///
    ///     // Gets a block by its hash with full transactions.
    ///     let block = provider.get_block_by_hash(block_hash).full().await.unwrap();
```

ZanCorDX (approved):

PR #519, opened 2025-03-26
New bundle version v2

ZanCorDX:
## 📝 Summary

A version field is added to bundles.
New uuid is used for bundles v2.

* [X] Run `make lint`
* [X] Run `make test`


crates/rbuilder/src/primitives/serialize.rs
```diff
@@ -1,8 +1,9 @@
 use super::{
```
dvush, line 1: Please add a test where version is not set in json.

dvush (approved):

PR #522, opened 2025-03-26
chore: fix realpath usage by removing unsupported `-s` flag

mdqst:
## 📝 Summary
I’ve fixed an issue with the `realpath` command in the script. The `-s` flag was causing errors on certain systems as it is not universally supported. I’ve removed the `-s` flag to ensure better compatibility, now using `realpath "$1"` and `realpath "$2"` for absolute paths.

## 💡 Motivation and Context
The `-s` flag was unsupported in some systems, causing errors. This change improves the script’s robustness and ensures it works properly across different environments.

---

* [x] Run `make lint`
* [x] Run `make test`


metachris:
thanks, but closing because not helpful as CI supports this.

PR #523, opened 2025-03-26
fix: payout gas limit estimation

efbig:
## 📝 Summary

Fix payout gas limit estimation for new revm semantics.

* [x] Run `make lint`
* [x] Run `make test`

### DIFF
--- a/crates/rbuilder/src/building/payout_tx.rs
+++ b/crates/rbuilder/src/building/payout_tx.rs
@@ -15,7 +15,7 @@ pub fn create_payout_tx(
     nonce: u64,
     to: Address,
     gas_limit: u64,
-    value: u128,
+    value: U256,
 ) -> Result<Recovered<TransactionSigned>, secp256k1::Error> {
     let tx = Transaction::Eip1559(TxEip1559 {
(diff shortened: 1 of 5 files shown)


crates/rbuilder/src/building/payout_tx.rs
```diff
@@ -53,22 +53,23 @@ pub fn insert_test_payout_tx(
...
         gas_limit,
-        0xFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFF,
+        U256::from(tx_value),
     )?;
 
     let mut db = state.new_db_ref();
     let mut evm = ctx.evm_factory.create_evm(db.as_mut(), ctx.evm_env.clone());
+
+    let cache_account = evm.db_mut().load_cache_account(builder_signer.address)?;
+    cache_account.increment_balance(tx_value * 2); // double to cover tx value and fee
```
dvush, line 71: What would happen if account balance is 0 and tx value is really slow?
We simulate top of block with balance is 0 (random coinbase) and this can break some refund bundles.

I think its safer to have value + some big value like base fee * 30mil

dvush (approved):

PR #524, opened 2025-03-27
refactor: Block Builder Executor

avalonche:
## 📝 Summary

op-rbuilder changes for reth 1.3.4. Reth had some major upheavals to its payload builder and generator and this mimics the op payload builder implementation as closely as possible.

* Introduction of Block Builder Executor, which means we need to implement our own custom executor to add revert protection
* Refactor of how payload builder is implemented in reth
* Clean up on the signer struct
* Note: flashblocks builder not in scope for this refactor
---

* [x] Run `make lint`
* [x] Run `make test`

### DIFF
--- a/crates/rbuilder/src/live_builder/mod.rs
+++ b/crates/rbuilder/src/live_builder/mod.rs
@@ -361,11 +361,7 @@ where
     // used in the optimism context. If further customisation is required in the future
     // this should be improved on.
     fn timings(&self) -> TimingsConfig {
-        if cfg!(feature = "optimism") {
-            TimingsConfig::optimism()
-        } else {
-            TimingsConfig::ethereum()
-        }
+        TimingsConfig::ethereum()
     }
 }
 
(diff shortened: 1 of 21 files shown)


crates/rbuilder/src/live_builder/mod.rs
```diff
@@ -361,11 +361,7 @@ where
     // used in the optimism context. If further customisation is required in the future
     // this should be improved on.
     fn timings(&self) -> TimingsConfig {
-        if cfg!(feature = "optimism") {
-            TimingsConfig::optimism()
-        } else {
-            TimingsConfig::ethereum()
-        }
+        TimingsConfig::ethereum()
```
ZanCorDX, line 364: Is the comment still valid or did we deprecate some optimist compatibility?

avalonche: this was left behind by Liam, when we used the live builder from rbuilder for op-rbuilder. All optimism code in rbuilder should not be needed now

ZanCorDX: Can you remove the comment then?

PR #543, opened 2025-04-07
Add metrics for flashblock and message tracking in OpRBuilder

jowparks:
Enhanced the OpRBuilderMetrics struct to include new metrics for tracking flashblocks, messages sent, and invalid blocks. Updated the OpPayloadBuilder to utilize these metrics during block building and transaction simulation processes.

- Added `flashblock_count`, `messages_sent_count`, and `invalid_blocks_count` to OpRBuilderMetrics.
- Integrated metrics tracking in OpPayloadBuilder for message sending, block building success, and transaction simulation.
- Recorded durations for various operations to improve performance monitoring.

## 💡 Motivation and Context

---

* [ ] Run `make lint`
* [ ] Run `make test`

### DIFF
--- a/crates/op-rbuilder/src/payload_builder.rs
+++ b/crates/op-rbuilder/src/payload_builder.rs
@@ -417,7 +417,6 @@ where
             self.metrics
                 .total_block_built_duration
                 .record(block_build_start_time.elapsed());
-            self.metrics.block_built_success.increment(1);
 
             // return early since we don't need to build a block with transactions from the pool
             return Ok(());
@@ -489,6 +488,7 @@ where
                         flashblock_count,
                     );
 
+                    let flashblock_build_start_time = Instant::now();
                     let state = StateProviderDatabase::new(&state_provider);
 
                     let mut db = State::builder()
@@ -526,11 +526,11 @@ where
                         return Ok(());
                     }
 
-                    let state_merge_start_time = Instant::now();
+                    let total_block_built_duration = Instant::now();
                     let build_result = build_block(db, &ctx, &mut info);
                     ctx.metrics
-                        .state_transition_merge_duration
-                        .record(state_merge_start_time.elapsed());
+                        .total_block_built_duration
+                        .record(total_block_built_duration.elapsed());
 
                     // Handle build errors with match pattern
                     match build_result {
@@ -551,6 +551,10 @@ where
                                 error!(target: "payload_builder", "Failed to send flashblock message: {}", err);
                             }
 
+                            // Record flashblock build duration
+                            self.metrics
+                                .flashblock_build_duration
+                                .record(flashblock_build_start_time.elapsed());
                             ctx.metrics
                                 .payload_byte_size
                                 .record(new_payload.block().size() as f64);
@@ -569,9 +573,6 @@ where
                 }
                 None => {
                     // Exit loop if channel closed or cancelled
-                    self.metrics
-                        .total_block_built_duration
-                        .record(block_build_start_time.elapsed());
                     self.metrics.block_built_success.increment(1);
                     self.metrics
                         .flashblock_count
--- a/crates/op-rbuilder/src/payload_builder.rs
+++ b/crates/op-rbuilder/src/payload_builder.rs
@@ -324,8 +324,9 @@ where
 {
     /// Send a message to be published
     pub fn send_message(&self, message: String) -> Result<(), Box<dyn std::error::Error>> {
+        self.tx.send(message)?;
         self.metrics.messages_sent_count.increment(1);
-        self.tx.send(message).map_err(|e| e.into())
+        Ok(())
     }
 
     /// Constructs an Optimism payload from the transactions sent via the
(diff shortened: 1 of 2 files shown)


crates/op-rbuilder/src/payload_builder.rs
```diff
@@ -47,7 +48,7 @@ use reth_payload_builder_primitives::PayloadBuilderError;
 use reth_payload_primitives::PayloadBuilderAttributes;
 use reth_payload_util::{BestPayloadTransactions, PayloadTransactions};
 use reth_primitives::{BlockBody, SealedHeader};
-use reth_primitives_traits::{proofs, Block as _, SignedTransaction};
+use reth_primitives_traits::{proofs, Block as _, InMemorySize, SignedTransaction};
```
cody-wang-cb, line 51: curious what's `InMemorySize` used for, doesn't seem like it's used anywhere

crates/op-rbuilder/src/payload_builder.rs
```diff
@@ -498,23 +526,56 @@ where
...
-
-                    best_payload.set(new_payload.clone());
-                    bundle_state = new_bundle_state;
-                    total_gas_per_batch += gas_per_batch;
-                    flashblock_count += 1;
+                    let state_merge_start_time = Instant::now();
+                    let build_result = build_block(db, &ctx, &mut info);
+                    ctx.metrics
+                        .state_transition_merge_duration
+                        .record(state_merge_start_time.elapsed());
```
cody-wang-cb: did you mean to use another metric here? Seems duplicated

crates/op-rbuilder/src/payload_builder.rs
```diff
@@ -374,31 +380,45 @@ where
...
         if ctx.attributes().no_tx_pool {
             tracing::info!(
                 target: "payload_builder",
                 "No transaction pool, skipping transaction pool processing",
             );
 
+            self.metrics
+                .total_block_built_duration
+                .record(block_build_start_time.elapsed());
+            self.metrics.block_built_success.increment(1);
```
cody-wang-cb: should probably not use metrics here because when there's `no_tx_pool` it means it's just syncing, not building blocks

crates/op-rbuilder/src/payload_builder.rs
```diff
@@ -498,23 +526,56 @@ where
...
+                            total_gas_per_batch += gas_per_batch;
+                            flashblock_count += 1;
+                            tracing::info!(target: "payload_builder", "Flashblock {} built", flashblock_count);
+                        }
+                    }
                 }
                 None => {
                     // Exit loop if channel closed or cancelled
+                    self.metrics
+                        .total_block_built_duration
```
cody-wang-cb: In addition to total block built duration, it'd be good to track individual flashblock duration as well

crates/op-rbuilder/src/payload_builder.rs
```diff
@@ -1126,6 +1219,19 @@ where
             info.executed_transactions.push(tx.into_inner());
         }
 
+        self.metrics
```
cody-wang-cb, line 1224: Nice!

crates/op-rbuilder/src/payload_builder.rs
jowparks: In rust traits can be implicitly used, here it is used in:
`self.metrics.tx_byte_size.record(tx.inner().size() as f64);`
Where the `.size()` is implemented here:
```
impl InMemorySize for OpTransactionSigned {
    #[inline]
    fn size(&self) -> usize {
        mem::size_of::<TxHash>() + self.transaction.size() + mem::size_of::<Signature>()
    }
}
```

crates/op-rbuilder/src/payload_builder.rs
jowparks: Good call!

crates/op-rbuilder/src/payload_builder.rs
jowparks: `total_block_built_duration`

crates/op-rbuilder/src/payload_builder.rs
```diff
@@ -320,6 +324,7 @@ where
 {
     /// Send a message to be published
     pub fn send_message(&self, message: String) -> Result<(), Box<dyn std::error::Error>> {
+        self.metrics.messages_sent_count.increment(1);
```
avalonche, line 328: should this be recorded after the message is sent? this way if the `send` message errors the counter won't increment

crates/op-rbuilder/src/payload_builder.rs
```diff
@@ -498,23 +526,57 @@ where
...
+                    ctx.metrics
+                        .total_block_built_duration
+                        .record(total_block_built_duration.elapsed());
+
+                    // Handle build errors with match pattern
+                    match build_result {
+                        Err(err) => {
+                            // Track invalid/bad block
+                            self.metrics.invalid_blocks_count.increment(1);
+                            error!(target: "payload_builder", "Failed to build flashblock {}: {}", flashblock_count, err);
```
avalonche: it would be useful to include the block number here too in the logging

avalonche (approved):

PR #544, opened 2025-04-08
Introduce ThreadBlockBuildingContext, remove old CachedReads.

dvush:
The idea behind ThreadBlockBuildingContext is to have a place that is owned by a particular thread (for example, max profit ordering builder) that can be used as a scratchpad for caching without doing global caching shared between all threads.

Currently we have only one place for caches BlockBuildingContext. It is global and shared between all builders, top of block simulation, and finalization thread. This approach solves many problems with caching. For example, we use it for eth-sparse-mpt root hash caching.

But sometimes we would like to have a cache that is local to the current thread to avoid the overhead of mutex and multiple threads.

I've seen the need for caches like this multiple times. This commit adds support for local caches like this.

The first thing that is moved to this new setup of 2 caches is cached reads. We had a lot of trouble with passing CachedReads struct around and it introduced itself into traits where it does not belong such as BlockBuildingHelper or backtesting code. Here we move CachedReads to this uniform setup and it simplifies cached reads state handling.

## 💡 Motivation and Context

---

* [ ] Run `make lint`
* [ ] Run `make test`

### DIFF
--- a/crates/rbuilder/src/backtest/build_block/backtest_build_block.rs
+++ b/crates/rbuilder/src/backtest/build_block/backtest_build_block.rs
@@ -117,14 +117,13 @@ where
                     builder_name: builder_name.clone(),
                     sim_orders: &sim_orders,
                     provider: provider_factory.clone(),
-                    cached_reads: None,
                 };
                 let build_res = config.build_backtest_block(builder_name, input);
                 if let Err(err) = &build_res {
                     println!("Error building block: {:?}", err);
                     return None;
                 }
-                let (block, _) = build_res.ok()?;
+                let block = build_res.ok()?;
                 println!(
                     "Built block {} with builder: {:?}",
                     ctx.block(),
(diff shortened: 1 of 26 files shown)


ZanCorDX:
Is it possible to hide the ThreadBlockBuildingContext inside the BlockBuildingHelperFromProvider so we don't have to pass it to edit the block since it's always the same?

dvush:
I decided against it because we move block building helper to send to other thread for sealing. This would be against the idea of having fixed cache that is:
* tied to some thread 
* never cloned during the slot
* it only fills with data and grows

If you put thread local cache with block building helper that would mean that we need to clone it every time we finish block building attempt.

dvush:
The only way I see to make it work would be putting `Mutex<Option<Arc<ThreadLocalBuildingContext>>>` into block building helper and adding the following flow: you set up context for the context of the current thread and then you remove it when you send it to other thread. This is error prone and ugly, passing `&mut ` to every method that requires it seems to be much better and you are actually protected by borrow checker since `&mut` is correct way to have this context in your code.

ZanCorDX (approved):

PR #554, opened 2025-04-11
Feat: `evm` module with custom `EvmFactory` trait

julio4:
## 💡 Motivation and Context

This PR brings small code cleanup for `EvmFactory`. I have created a new `EvmFactory` trait in rbuilder because the `Evm` trait from `alloy_revm` has a lot of associated types and trait bounds.
These are now defined in the custom `EvmFactory` trait and resolved in the implementation for `EthCachedEvmFactory`, so this removes the need for extra trait bounds outside of the `evm` module (such as in the `execute_evm` function in `order_commit`).

## 📝 Summary

- The `EthCachedEvmFactory` was moved to new module `evm` under `building` (`rbuilder::building::evm`), IMO this makes the overall code structure better as the separation of concerns is cleaner (more obvious)
- `EthCachedEvmFactory::Context` is defined to `EthEvmContext<DB>`, which is an alias of `Context<BlockEnv, TxEnv, CfgEnv, DB>`
- `EvmFactory` custom trait, which is a more concrete version than `alloy_revm::EvmFactory`
- Remove the dependency on `EthEvmFactory` in `EthCachedEvmFactory` struct, but still use it's default implementation under the hood (we can also consider completely remove it and use `Context` instead)

---

* [x] Run `make lint`
* [x] Run `make test`

### DIFF
--- a/crates/rbuilder/src/building/evm.rs
+++ b/crates/rbuilder/src/building/evm.rs
@@ -1,0 +1,77 @@
+use crate::building::precompile_cache::{PrecompileCache, WrappedPrecompile};
+use parking_lot::Mutex;
+use reth_evm::{eth::EthEvmContext, EthEvm, EthEvmFactory, EvmEnv, EvmFactory};
+use revm::{
+    context::{
+        result::{EVMError, HaltReason},
+        TxEnv,
+    },
+    handler::EthPrecompiles,
+    inspector::NoOpInspector,
+    interpreter::interpreter::EthInterpreter,
+    primitives::hardfork::SpecId,
+    Database, Inspector,
+};
+use std::sync::Arc;
+
+#[derive(Debug, Clone, Default)]
+pub struct EthCachedEvmFactory {
+    evm_factory: EthEvmFactory,
+    cache: Arc<Mutex<PrecompileCache>>,
+}
+
+impl EvmFactory for EthCachedEvmFactory {
+    type Evm<DB, I>
+        = EthEvm<DB, I, WrappedPrecompile<EthPrecompiles>>
+    where
+        DB: Database<Error: Send + Sync + 'static>,
+        I: Inspector<EthEvmContext<DB>>;
+
+    type Context<DB>
+        = EthEvmContext<DB>
+    where
+        DB: Database<Error: Send + Sync + 'static>;
+
+    type Error<DBError>
+        = EVMError<DBError>
+    where
+        DBError: core::error::Error + Send + Sync + 'static;
+
+    type Tx = TxEnv;
+    type HaltReason = HaltReason;
+    type Spec = SpecId;
+
+    fn create_evm<DB>(&self, db: DB, input: EvmEnv) -> Self::Evm<DB, NoOpInspector>
+    where
+        DB: Database<Error: Send + Sync + 'static>,
+    {
+        let evm = self
+            .evm_factory
+            .create_evm(db, input)
+            .into_inner()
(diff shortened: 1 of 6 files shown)


ZanCorDX:
Please add comments on trait EvmFactory (reason of existing, basically what you explain here!).

ZanCorDX:
Changing the dependency on EthEvmFactory and calling EthEvmFactory::default() seems dangerous, you are trusting the implementation of EthEvmFactory will not change. If in the future EthEvmFactory::default() gets super expensive and stores some internal data we'll get a performance penalty.

julio4:
> Changing the dependency on EthEvmFactory and calling EthEvmFactory::default() seems dangerous, you are trusting the implementation of EthEvmFactory will not change. If in the future EthEvmFactory::default() gets super expensive and stores some internal data we'll get a performance penalty.

I see, because before it was calling `EthEvmFactory::default()` at BlockBuildingContext not at each `create_evm` calls.
I can use `Context` instead of `EthEvmFactory::default()`, [by reverting this commit](https://github.com/flashbots/rbuilder/pull/554/commits/1052c799c2483a41462344424919a0527ba086e5).

> Please add comments on trait EvmFactory (reason of existing, basically what you explain here!).

Done.

ZanCorDX:
> > Changing the dependency on EthEvmFactory and calling EthEvmFactory::default() seems dangerous, you are trusting the implementation of EthEvmFactory will not change. If in the future EthEvmFactory::default() gets super expensive and stores some internal data we'll get a performance penalty.
> 
> I see, because before it was calling `EthEvmFactory::default()` at BlockBuildingContext not at each `create_evm` calls. I can use `Context` instead of `EthEvmFactory::default()`, [by reverting this commit](https://github.com/flashbots/rbuilder/pull/554/commits/1052c799c2483a41462344424919a0527ba086e5).
> 
> > Please add comments on trait EvmFactory (reason of existing, basically what you explain here!).
> 
> Done.

You are copy/pasting the code of EthEvmFactory. Unless you can show me something we win with the copy/paste, can we go back to having a EthEvmFactory inside EthCachedEvmFactory?

julio4:
> You are copy/pasting the code of EthEvmFactory. Unless you can show me something we win with the copy/paste, can we go back to having a EthEvmFactory inside EthCachedEvmFactory?

Yes this is done one purpose to have the same logic as `EthEvmFactory` without having to trust that its implementation will not change in the future.
If you think this is not a good idea, I can still revert to having the `evm_factory` field in `EthCachedEvmFactory`, and only call `EthEvmFactory::Default()` within `EthCachedEvmFactory::Default()`.

ZanCorDX:
> > You are copy/pasting the code of EthEvmFactory. Unless you can show me something we win with the copy/paste, can we go back to having a EthEvmFactory inside EthCachedEvmFactory?
> 
> Yes this is done one purpose to have the same logic as `EthEvmFactory` without having to trust that its implementation will not change in the future. If you think this is not a good idea, I can still revert to having the `evm_factory` field in `EthCachedEvmFactory`, and only call `EthEvmFactory::Default()` within `EthCachedEvmFactory::Default()`.

Yes please 🙏 I trust reth people, if they decide to change EthEvmFactory it will be for good reasons.

julio4:
> Yes please 🙏 I trust reth people, if they decide to change EthEvmFactory it will be for good reasons.

Done. Thanks for the valuable comments.

ZanCorDX (approved):

PR #555, opened 2025-04-11
Feat: new PR template proposal

julio4:
## 💡 Motivation and Context

This PR moves the "Motivation and Context" section to the top of the PR Template. IMO this order encourages contributors to explain the "why" before the "what", making the flow more intuitive for both authors and reviewers.
"Summary" section was also marked as optional because for smaller PRs it can easily be inferred from the diff.

### DIFF
--- a/.github/pull_request_template.md
+++ b/.github/pull_request_template.md
@@ -1,10 +1,11 @@
-## 📝 Summary
-
-<!--- A general summary of your changes -->
-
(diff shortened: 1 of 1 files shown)


ZanCorDX:
The person in change of this stuff is going on vacation, this might take some time to be reviewed.

metachris:
Discussed, and we prefer summary first. Thanks nevertheless.

PR #560, opened 2025-04-14
CompetitionBid is now a struct instead of a plain U256

ZanCorDX:
## 📝 Summary

This change changes the bid U256 in BidValueObs for a struct:
```
#[derive(Clone, Debug)]
pub struct CompetitionBid {
    bid: U256,
    /// For metrics. Set on creation which is the first time we see it in our process.
    creation_time: OffsetDateTime,
}
```

## 💡 Motivation and Context

This change will allow better metrics on time to response for competition bids.

* [X] Run `make lint`
* [X] Run `make test`

### DIFF
--- a/crates/rbuilder/src/live_builder/block_output/bid_value_source/best_bid_sync_source.rs
+++ b/crates/rbuilder/src/live_builder/block_output/bid_value_source/best_bid_sync_source.rs
@@ -1,4 +1,4 @@
-use super::interfaces::{BidValueObs, BidValueSource};
+use super::interfaces::{BidValueObs, BidValueSource, CompetitionBid};
 use alloy_primitives::U256;
 use parking_lot::Mutex;
 use std::sync::Arc;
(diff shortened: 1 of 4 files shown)


crates/rbuilder/src/live_builder/block_output/bid_value_source/best_bid_sync_source.rs
```diff
@@ -31,17 +31,21 @@ impl BestBidSyncSource {
...
 #[derive(Debug, Default)]
 struct BestBidSyncSourceInner {
-    best_bid: Mutex<Option<U256>>,
+    best_bid: Mutex<Option<CompetitionBid>>,
 }
 
 impl BidValueObs for BestBidSyncSourceInner {
-    fn update_new_bid(&self, bid: U256) {
+    fn update_new_bid(&self, bid: CompetitionBid) {
         *self.best_bid.lock() = Some(bid);
```
avalonche, line 49: when the bid is updated, do you need to update the creation time?

ZanCorDX: Yes, the new timestamp is there for metrics, represents when we saw it at the source so it must be propagated so at some point we can measure our response time.
But anyway BestBidSyncSourceInner is a helper, not really used in the bidding circuit.

avalonche (approved):

PR #567, opened 2025-04-16
fix: don't build flashblocks with more gas than block gas limit

danyalprout:
## 📝 Summary

Currently it's possible for the builder to make more Flashblocks than intended during a block (likely due to delays). If more flashblocks are made than desired (e.g. 11 flashblocks for a 2s block w/ 200ms Flashblocks), the 11th flashblock will have a gas limit that is over the blocks gas limit and will cause invalid blocks on the sequencer.

## 💡 Motivation and Context

If the builder produces a full block that uses more gas than the gas limit, the local EL client will treat it as invalid.

---

* [ ] Run `make lint`
* [x] Run `make test`

### DIFF
--- a/crates/op-rbuilder/src/payload_builder.rs
+++ b/crates/op-rbuilder/src/payload_builder.rs
@@ -232,6 +232,8 @@ pub struct OpPayloadBuilder<Pool, Client> {
     pub chain_block_time: u64,
     /// Flashblock block time
     pub flashblock_block_time: u64,
+    /// Number of flashblocks per block
+    pub flashblocks_per_block: u64,
     /// The metrics for the builder
     pub metrics: OpRBuilderMetrics,
 }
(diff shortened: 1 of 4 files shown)


crates/op-rbuilder/src/payload_builder.rs
```diff
@@ -512,7 +526,7 @@ where
                         &mut info,
                         &mut db,
                         best_txs,
-                        total_gas_per_batch,
+                        total_gas_per_batch.min(ctx.block_gas_limit()),
```
danyalprout, line 529: This technically isn't necessary given we skip building the flashblock if it's out of the expected bounds. However this seems like a sensible check to keep IMO

avalonche (approved):

danyalprout:
@avalonche would it be possible to run the CI please 🙏

PR #568, opened 2025-04-17
Fast relays cfg

ZanCorDX:
## 📝 Summary

Added a 2 new cfg:
- Relay cfg has:
    /// critical blocks go only to fast relays. None -> true
    pub is_fast: Option<bool>,
- L1Config:
    /// Bids above this value will only go to fast relays.
    pub fast_bid_threshold_eth: String,
A block is sent to not fast relays only if its bid is < fast_bid_threshold_eth and it contains no bundle with replacement id.

* [X] Run `make lint`
* [X] Run `make test`


dvush (approved):

metachris:
See also https://collective.flashbots.net/t/updated-mev-boost-relay-settings/4881

PR #576, opened 2025-04-23
Add metric to track number of invalid removed transactions

SozinM:
## 📝 Summary

Nice addition to track how much invalid transaction we remove during block building

## 💡 Motivation and Context

---

* [ ] Run `make lint`
* [ ] Run `make test`


SozinM:
realized that there is already payload_num_tx_simulated_fail that reflect number of reverted transactions

PR #579, opened 2025-04-23
record metrics on cancel

jowparks:
## 💡 Motivation and Context

---

* [ ] Run `make lint`
* [ ] Run `make test`


avalonche (approved):
a more robust implementation would be having the metrics returned from `pub fn execute_best_transactions<DB>(` and record the metrics when this function is called without considering all the extra code branches.

PR #595, opened 2025-04-29
Add flashblocks feature to ci

SozinM:
## 📝 Summary

We need flashblocks feature to build docker images

## 💡 Motivation and Context

---

* [ ] Run `make lint`
* [ ] Run `make test`

### DIFF
--- a/.github/workflows/release-op.yaml
+++ b/.github/workflows/release-op.yaml
@@ -27,6 +27,7 @@ on:
         options:
           - ''
           - 'redact-sensitive'
+          - 'flashblocks'
         required: false
         type: choice
 


avalonche (commented):
we should have two workflows - one for the regular op-rbuilder and flashblocks op-rbuilder

SozinM:
Afaik only difference is the feature flash, is there something else @avalonche?

avalonche:
adding the feature flag will result in a different binary. with the flashblocks feature you can not change back to the regular payload builder via config / cmd line and need to rebuild the image

avalonche (approved):

PR #605, opened 2025-04-30
Fast relay concept is split between fast and independent

ZanCorDX:
## 📝 Summary

- new flag "is_independent" for relays.
- "fast_bid_threshold_eth" renamed to "independent_bid_threshold_eth"
- Critical blocks (containing bundles with replacement ids) go only to fast relays. None -> true
- Big blocks (bid > independent_bid_threshold_eth) go only to independent relays. None -> true

* [X] Run `make lint`
* [X] Run `make test`

### DIFF
--- a/crates/rbuilder/src/live_builder/block_output/relay_submit.rs
+++ b/crates/rbuilder/src/live_builder/block_output/relay_submit.rs
@@ -370,6 +370,9 @@ fn submit_block_to_relays(
 }
 
 /// Creates a Fn to decide if the block should go to a relay.
+/// The cfg defines 2 flags on relays: fast and independent.
+/// If a block has replacement ids it should NOT go to a relay that is not fast since it needs fast cancellations.
+/// If a block is expensive it should NOT go to a non independent relay.
 fn get_relay_filter_and_update_metrics(
     block: &Block,
     optimistic: bool,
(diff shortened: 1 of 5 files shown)


crates/rbuilder/src/live_builder/block_output/relay_submit.rs
```diff
@@ -378,8 +388,16 @@ fn can_send_to_slow_relay(block: &Block, fast_bid_threshold: U256) -> bool {
             Order::Tx(_) => false,
             Order::ShareBundle(_) => false,
         });
-    let is_expensive_block = block.trace.bid_value > fast_bid_threshold;
-    !has_replacement_uuid && !is_expensive_block
+    inc_initiated_submissions(optimistic, !only_fast, !only_independent);
+    move |relay: &MevBoostRelayBidSubmitter| {
+        if only_independent && !relay.is_independent() {
```
dvush, line 396: I don't fully understand the logic from the description and the docs and this code. This means the description can be improved.  

Relays can be independent and fast at the same time?

* only_fast bool is practically going to be true for every block
* only_independent will be true for big blocks

So what will happen for big blocks with replacements?

They will be sent only to fast AND independent relays. Is this intended behavior?

dvush (changes requested):

crates/rbuilder/src/live_builder/block_output/relay_submit.rs
```diff
@@ -378,8 +388,16 @@ fn can_send_to_slow_relay(block: &Block, fast_bid_threshold: U256) -> bool {
             Order::Tx(_) => false,
             Order::ShareBundle(_) => false,
```
dvush, line 392: share bundles can have replacements too

dvush:
here is a thread with explanations https://collective.flashbots.net/t/updated-mev-boost-relay-settings/4881

dvush (approved):
I don't have issues with implementation, only the docs, so I'll approve.

crates/rbuilder/src/live_builder/block_output/relay_submit.rs
ZanCorDX: Updated the comments.

PR #606, opened 2025-05-01
Finish block building process even when cancel request is found

ferranbt:
## 📝 Summary

This PR introduces three changes:
- It refactors some of the common utilities shared among the tests into a single `TestHarness` struct which spins the test framework with all the components. Note that it does not update the tests to use this new utility.
- It fixes an issue with the block builder that would stop the block building process and not return any block if a cancel request was found. This happens when an FCU and a getPayload request are called to close to each other, the getPayload cancels the block building process, and getPayload waits forever for a block that will never be built. Now, the block building finishes.
- It adds an integration test to cover this use case with the new utility.

## 💡 Motivation and Context

---

* [ ] Run `make lint`
* [ ] Run `make test`

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -7469,6 +7469,7 @@ dependencies = [
  "op-alloy-network",
  "op-alloy-rpc-types-engine",
  "op-revm",
+ "parking_lot",
  "rand 0.9.0",
  "reth",
  "reth-basic-payload-builder",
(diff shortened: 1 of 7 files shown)


avalonche:
>  the getPayload cancels the block building process, and getPayload waits forever for a block that will never be built

this would only happen if fcu causes a cancel before getPayload is resolved right? why would getPayload cause a cancel?

avalonche (approved):

Issue #613, opened 2025-05-07
Respect Max DA Sizes


danyalprout:
The OP Stack currently has two parameters that are used to throttle transaction inclusion:

**max da tx size**: 
The maximum size of a fastlz compressed transaction ([one](https://github.com/ethereum-optimism/op-geth/blob/optimism/core/txpool/legacypool/legacypool.go#L593), [two](https://github.com/ethereum-optimism/op-geth/blob/optimism/core/types/rollup_cost.go#L563), [three](https://github.com/ethereum-optimism/op-geth/blob/optimism/core/types/rollup_cost.go#L551-L559))
**max da block size**: 
Only include transactions where the sum of the fastlz size is less than the maxda block size ([one](https://github.com/ethereum-optimism/op-geth/blob/optimism/miner/worker.go#L543-L556), [two](https://github.com/ethereum-optimism/op-geth/blob/optimism/miner/worker.go#L543-L556))

Currently rbuilder has two payload builders (vanilla + flashblocks enabled), which have different behavior to op-geth sequencer:

* vanilla
  * a different builder config is used between the RPC endpoint and the builder, so I think the settings will never be applied in the builder (due to creating a separate instance [here](https://github.com/flashbots/rbuilder/blob/develop/crates/op-rbuilder/src/payload_builder_vanilla.rs#L243))
  * max da tx/block size use the size of the transaction (non compressed) ([one](https://github.com/flashbots/rbuilder/blob/develop/crates/op-rbuilder/src/primitives/reth/execution.rs#L65-L71))
* flashblocks
  * max da tx size is not supported
  * max da block size is not supported

I can take a look at this just wanted to align on the changes (cc @ferranbt)

- [x] Vanilla builder
    - [x] Share the same underlying builder config so params are passed from the RPC call to the builder ([PR](https://github.com/flashbots/rbuilder/pull/616))
    - [x] Switch `is_tx_over_limits` to use the fastlz txn size ([PR](https://github.com/flashbots/rbuilder/pull/615))
    - [x] Isn't currently incrementing culmative da bytes used ([PR](https://github.com/flashbots/rbuilder/pull/615))
- [x] Flashblocks builder
    - [x] Pass through the builder config ([PR](https://github.com/flashbots/rbuilder/pull/616))
    - [x] Add tx size filtering (using fastlz) ([PR](https://github.com/flashbots/rbuilder/pull/616))
[...]

PR #615, opened 2025-05-08
fix: use fastlz sizes for tracking da usage

danyalprout:
## 📝 Summary

This PR has a couple of changes:

* Use the FastLZ size of transactions to verify per tx and per block DA limits. This matches the op-geth sequencer behaviour. See this [issue](https://github.com/flashbots/rbuilder/issues/613) for more details on how op-geth behaves. 
* Track cumulative da bytes used 

## 💡 Motivation and Context

Without this change:

* The DA limiting is based on the uncompressed transaction size, which will result in needlessly throttling very compressible transactions.
* Throttling does not take into account the amount of bytes currently being used in the block

---

* [x] Run `make lint`
* [x] Run `make test`

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -7466,6 +7466,7 @@ dependencies = [
  "jsonrpsee 0.24.9",
  "metrics",
  "op-alloy-consensus",
+ "op-alloy-flz",
  "op-alloy-network",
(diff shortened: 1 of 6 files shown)


danyalprout:
Closing in favor of https://github.com/flashbots/op-rbuilder/pull/7

PR #616, opened 2025-05-08
fix: ensure DA transaction throttling uses params that are set

### LINKED ISSUE #613: Respect Max DA Sizes

danyalprout:
## 📝 Summary

* Ensure the builders share the same `OpDAConfig` that the RPC endpoint does. 
* For Flashblocks builder, consider the DA param values

## 💡 Motivation and Context

Prior to this PR, I believe setting throttling via the RPC would not have any impact as an unrelated DAConfig object is created [here](https://github.com/flashbots/rbuilder/blob/develop/crates/op-rbuilder/src/payload_builder_vanilla.rs#L243). Part of the work to fix: https://github.com/flashbots/rbuilder/issues/613

---

* [x] Run `make lint`
* [x] Run `make test`

### DIFF
--- a/crates/op-rbuilder/src/integration/integration_test.rs
+++ b/crates/op-rbuilder/src/integration/integration_test.rs
@@ -438,7 +438,9 @@ mod tests {
             .await?;
 
         // check there's 10 flashblocks log lines (2000ms / 200ms)
-        op_rbuilder.find_log_line("Building flashblock 9").await?;
+        op_rbuilder
+            .find_log_line("Building flashblock idx=9")
+            .await?;
 
         // Process websocket messages
         let timeout_duration = Duration::from_secs(10);
@@ -571,7 +573,9 @@ mod tests {
             .await?;
 
         // check there's no more than 10 flashblocks log lines (2000ms / 200ms)
-        op_rbuilder.find_log_line("Building flashblock 9").await?;
+        op_rbuilder
+            .find_log_line("Building flashblock idx=9")
+            .await?;
         op_rbuilder
             .find_log_line("Skipping flashblock reached target=10 idx=10")
             .await?;
--- a/crates/op-rbuilder/src/main.rs
+++ b/crates/op-rbuilder/src/main.rs
@@ -1,14 +1,14 @@
 use clap::Parser;
 use monitoring::Monitoring;
-use reth::providers::CanonStateSubscriptions;
-use reth_optimism_cli::{chainspec::OpChainSpecParser, Cli};
-use reth_optimism_node::node::OpAddOnsBuilder;
-use reth_optimism_node::OpNode;
-
 #[cfg(feature = "flashblocks")]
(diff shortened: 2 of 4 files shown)


danyalprout:
Closing in favor of https://github.com/flashbots/op-rbuilder/pull/8

PR #620, opened 2025-05-15
Refactoring unifying tx+receipt+gas into a struct TransactionExecutionInfo.

ZanCorDX:
## 📝 Summary

In a few places we had a vector of tx and a brother vector of receipt matching one the txs.
I put both in a new struct TransactionExecutionInfo and I also moved there gas and added profit.
Some copy paste was removed.

## 💡 Motivation and Context

Nicer code and it will allow better checks on bundle execution due to the new per tx profit info.

* [X] Run `make lint`
* [X] Run `make test`

### DIFF
--- a/crates/rbuilder/src/building/order_commit.rs
+++ b/crates/rbuilder/src/building/order_commit.rs
@@ -181,19 +181,25 @@ where
         &mut self.db
     }
 }
-
+/// Common data used by TransactionOk/BundleOk
+#[derive(Debug, Clone, PartialEq, Eq)]
+pub struct TransactionExecutionInfo {
+    pub tx: TransactionSignedEcRecoveredWithBlobs,
+    pub receipt: Receipt,
+    pub gas_used: u64,
+    /// On loss capped to 0
+    pub coinbase_profit: U256,
+}
 #[derive(Debug, Clone)]
 pub struct TransactionOk {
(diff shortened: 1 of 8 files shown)


crates/rbuilder/src/building/order_commit.rs
```diff
@@ -181,19 +181,25 @@ where
...
     }
 }
-
+/// Common data used by TransactionOk/BundleOk
+#[derive(Debug, Clone, PartialEq, Eq)]
+pub struct TransactionExecutionInfo {
+    pub tx: TransactionSignedEcRecoveredWithBlobs,
+    pub receipt: Receipt,
+    pub gas_used: u64,
+    /// On loss capped to 0
```
dvush: I thing this change may introduce bugs and issues in the future.

Before we only put coinbase profit for order as a whole but now we do it per tx and we don't count negative profits (such as  refund txs profits).

It was OK to put U256 to order coinbase profit because we disallow orders with negative profits overall so it would be impossible to have negative value there on the other hand its completely fine to have txs with negative profits.

I think it should be I256 here because this type is more fit here than 256. IIRC we have some methods in utils to work with u256/i256 since its pretty underdeveloped in alloy.

dvush (approved):

PR #621, opened 2025-05-16
Non mempool profit

ZanCorDX:
## 📝 Summary

SimValue now has 2 ProfitInfo instead of the single coinbase_profit/gas_used pair so now we have the full_profit_info (previous coinbase_profit/gas_used) and the new non_mempool_profit_info which excludes mempool txs profits from bundles.
Backwards compatible.
Implementation to detect mempool tx is super simple and heavily assumes that we will always see a mempool tx before a bundle including that tx:
We add a order_input::mempool_txs_detector::ReplaceableOrderStreamSniffer in the orderflow stream that updates the MempoolTxsDetector which is used on SimValue creation to filter mempool txs.

## 💡 Motivation and Context

This will give more info to block building algorithms.

* [X] Run `make lint`
* [X] Run `make test`

### DIFF
--- a/crates/rbuilder/src/building/mod.rs
+++ b/crates/rbuilder/src/building/mod.rs
@@ -979,7 +979,7 @@ mod test {
     /// Create a bundle with 2 txs, one from mempool and the other not.
     /// sim_value.non_mempool_profit_info().coinbase_profit() should only sum the profit for the second.
     #[test]
-    fn test_create_sim_value_non_mempool_coinbase_profit() {
+    fn test_create_sim_value_bundle_non_mempool_coinbase_profit() {
         let detector = MempoolTxsDetector::new();
         let mut data_gen = TestDataGenerator::default();
         let tx1 = data_gen.create_tx_with_blobs_nonce(Default::default());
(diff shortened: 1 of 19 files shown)


crates/rbuilder/src/building/mod.rs
```diff
@@ -928,3 +931,142 @@ pub enum FillOrdersError {
...
+            .iter()
+            .filter(|tx_info| !mempool_detector.is_mempool(&tx_info.tx))
+            .map(|tx_info| tx_info.coinbase_profit)
+            .sum::<I256>();
+        if non_mempool_coinbase_profit.is_positive() {
+            non_mempool_coinbase_profit.unsigned_abs()
+        } else {
+            error!(
+            non_mempool_coinbase_profit = format_ether(non_mempool_coinbase_profit),
+            "Non mempool orders have always positive profit, how did we got a negative value????");
```
dvush: can you please rewrite the message to make it more boring and uniform

crates/rbuilder/src/live_builder/order_input/mempool_txs_detector.rs
```diff
@@ -0,0 +1,80 @@
...
+    fn is_alive(&self) -> bool {
+        self.sink.is_alive()
+    }
+}
+
+/// Given a TransactionSignedEcRecoveredWithBlobs answers if the tx is from the mempool or not.
+/// Current implementation is super simple, it just checks the tx hash against a set of hashes.
+#[derive(Debug)]
+pub struct MempoolTxsDetector {
+    mempool_txs: Mutex<HashSet<TxHash>>,
```
dvush: I think its always better to use something that has lower chance of blocking:

`DashSet` >  `RwLock<HashSet>` > `Mutex<HashSet>`

since we use dashmap here a lot, DashSet is a great choice (but don't forget to specify ahash as a hasher instead of default hasher)

dvush (approved):
besides 2 comments, IMO its fine

PR #634, opened 2025-05-27
Faster root hash and finalize

dvush:
# Summary
This PR introduces new implementation of eth_sparse_mpt (fast root hash for block building) and makes additional tweaks to make finalize step faster.

For staging builder metrics for root hash and finalize are as following:

```
root_hash:
p50: old - 4 ms new - 0.75 ms (5.3x speedup)
P90: old - 6.3 ms new - 1.4 ms (5.5x speedup)
p99: old - 12.5 ms new -  5 ms (2.5x speedup)
```

```
finalize:
p50: old - 4.6 ms new - 1.4 ms (3.3x speedup)
p90: old - 7.9 ms new - 2.4 ms (3.3x speedup)
p99: old - 15 ms new - 6.5 ms (2.3x speedup)
```
# Config

The new implementation is hidden behind 2 feature flags for now

```
root_hash_sparse_trie_version = "v2"
faster_finalize = true
```
Also I strongly suggest setting `RAYON_NUM_THREADS=4`  because having too many threads in the threadpool will reduce performance.

# Explanation

## Why its faster:
* New implementation is more efficient even without algorithmic changes as its removes all unnecessary memory allocations and stores data in arrays which makes memory layout more efficient. Benchmark for inserting and hashing 10k random values shows almost 2x speedup compared to previous version.
* There is a major algorithmic change in new implementation. Now root hash for block is calculated incrementally. We keep changed tries from previous calculations and only change and hash the diff of changes between previously built block and current block. This uses the fact that most of txs have the same execution from one block to the next one.
* In addition we use this new trie to calculate receipt and transactions root when `faster_finalize=true`. This implementation is faster even for this kind of work because it is parallel. But its important to control `RAYON_NUM_THREADS` because too many threads makes it slower.
* This does not change p100 performance and p99 speedup is smaller because those numbers are dominated by fetching proofs from the reth database.

## Why its important
This improves bidding performance significantly. When we bid we only change last tx in the block. Bidding case should have the same execution time as p50 and p90 values above. 

## Future work

Currently finalize is evenly divided between: root hash, header transaction root, header receipt root and memory drop of a block building helper.

Further improvement of finalize for the bidding case would require changes to the rest of block finalization part. 
* We can use incremental trie for receipt/tx roots (although I did not see a huge improvement in p50/p90 when I tried that)
* We should refactor block building pipeline and remove inefficiencies related to excessive allocations and memory layout.


ZanCorDX (approved):

crates/eth-sparse-mpt/Cargo.toml
```diff
@@ -36,15 +36,17 @@ alloy-trie.workspace = true
...
 parking_lot.workspace = true
+eyre = { workspace = true }
+arrayvec = "0.7.6"
+dashmap = { version = "6.1.0", features = ["rayon"] }
 
 # test deps
 hash-db = { version = "0.15.2", optional = true }
 triehash = { version = "0.8.4", optional = true }
 flate2 = { workspace = true, optional = true }
-eyre = { workspace = true, optional = true}
```
CasPerUD, line 44: -1

CasPerUD: 0000

Issue #638, opened 2025-06-02
Add fuzz test to validate the sparse merkle trie


ferranbt:
Run a fuzz test with the `sparse-merkle-trie` and `reth-trie` to validate the implementation

PR #641, opened 2025-06-05
Model testing

krl:
## 📝 Summary

This is an example on how to make model tests with quickcheck.

The keys being tested are of a fixed size, which seems to be a hidden requirement of at least the v2 structure. (see test `crash_example_v2` included, but ignored)

## 💡 Motivation and Context

Expanding test coverage with model testing.

## Todo

Actually also test the calculated root hashes.

### DIFF
--- a/crates/eth-sparse-mpt/src/model_test.rs
+++ b/crates/eth-sparse-mpt/src/model_test.rs
@@ -1,0 +1,69 @@
+use crate::v2::trie::Trie;
+use quickcheck::{quickcheck, Arbitrary, Gen};
+use std::collections::HashMap;
+
+// The maximum key size. keeping it relatively
+// small increases the chance of multiple
+// operations being executed against the same
+// key, which will tease out more bugs.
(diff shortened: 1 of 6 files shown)


krl:
Related to and fixes #638

crates/eth-sparse-mpt/src/model_test.rs
```diff
@@ -0,0 +1,150 @@
...
+#[derive(Clone, Copy, Debug, PartialEq, Eq, Hash)]
+struct FixedKey(FixedBytes<32>);
+
+impl FixedKey {
+    fn as_slice(&self) -> &[u8; 32] {
+        &self.0
+    }
+
+    fn from_string(s: &str) -> Self {
+        Self(keccak256("a"))
```
krl: FIXME: this should of course not be `"a"` here.

dvush:
Have you seen tests module? It's already doing that kind of test. It's using proptest though.

PR #652, opened 2025-07-02
New competition bid source.

ZanCorDX:
## 📝 Summary

The whole pipeline has been updated to use `BlockBid` (all bids seen in all relays) instead `CompetitionBid` (single best bid to beat).
This change gives more information to the bidding service and implies that the `SlotBidder` is now just a `UnfinishedBlockBuildingSink` and the `BiddingService` gets all the bids by implemening `BlockBidWithStatsObs`
```
pub trait BlockBidWithStatsObs: Send + Sync {
    /// Be careful, we don't assume any kind of filtering here so bid may contain our own bids.
    fn update_new_bid(&self, bid_with_stats: BlockBidWithStats);
}
```
and is his responsibility to let the active bidders know about the competition bids.
The rbuilder now connects to the bid-scraper to get those `BlockBid`.
Some other changes:
- ethers replaced by alloy
- Some refactoring on crates/rbuilder/src/live_builder/config.rs to simplify the creation of other cfgs.
- Some lint cleaning.

## 💡 Motivation and Context

Get rid of any external dependency to get competition bids.

* [X] Run `make lint`
* [X] Run `make test`


crates/bid-scraper/src/bid_scraper_client.rs
```diff
@@ -0,0 +1,69 @@
...
+    publisher_url: String,
+    timeout: Duration,
+    retry_wait: Duration,
+) {
+    let url = publisher_url.clone(); // for reuse in error handler
+    tokio::select! {
+        result = FutureRetry::new(
+            move || run_nng_subscriber(obs.clone(), publisher_url.clone(), timeout),
+            move |error: Box<dyn std::error::Error>| {
+                tracing::error!("Subscriber to {url} returned an error: {error:?}");
```
dvush: Pls change to error with args like we have elsewhere

dvush (approved):

PR #666, opened 2025-07-21
Support for Fusaka

bharath-123:
This PR updates rbuilder to support Fusaka. These changes have been tested locally using [Kurtosis](https://github.com/ethpandaops/ethereum-package)

Things to consider in rbuilder:

**Breaking changes:**
1. **alloy/nybbles:** https://github.com/alloy-rs/nybbles/pull/17 updates the internal representation of `Nibble` from a `SmallVec<[u8; 64]>` to a `U256`. This breaks code in the merkle eth sparse tree as the code indexes a lot into the `SmallVec` representation
2. **alloy/eips**: https://github.com/alloy-rs/alloy/blob/fusaka/devnet2/crates/eips/src/eip7594/sidecar.rs#L25 we need to use `BlobTransactionSidecarVariant` enum which contains PeerDas type blob sidecars too. The only different b/w EIP-7954 sidecars from EIP-4844 sidecars is EIP-7954 sidecars use `cell_proofs` instead of `proofs`. rbuilder just uses `BlobTransactionSidecar` which breaks with the fulu hard fork.
3. **revm/context**: https://github.com/bluealloy/revm/commit/c70bb9a8c3d2d1fa3f24bcba744d7825d80a7637 this commit updates the `BlockNumber` and `Timestamp` to use `U256` instead of `u64`. There are inconsistencies in code where u64 is used for block number which fails to build as a result.
4. **revm/state**: changes made for EIP-7907: https://github.com/bluealloy/revm/commit/d8c4567f34e57d7f19804fbe6bc8adf70ee30ace introduces the `CodeSize` field to the `AccountInfo` . It was previously an `Option` but the mentioned commit makes it a mandatory field. 
5. **alloy/evm**: The `Database` trait in the alloy/evm repo now implements `std::fmt::Debug` implemented by https://github.com/alloy-rs/evm/commit/3d6a01b3594f66aaf60bcc93b7b2e3cd552ec7a8. This broke code since a dependency was using the updated `Database` trait from alloy/evm. Using the `Database` trait from alloy/evm seems to have fixed this. 
6. **revm/inspector**: It appears that the `step` method has been removed from the `access_list_inspector`. We need to update the `revm-inspectors` crate.
7. The method `bytecode_by_hash` is now defined in trait `BytecodeReader` rather than in the `StateProvider`

**Notes**
1. We need to send `BlobsBundleV2` to the relay. https://github.com/alloy-rs/alloy/blob/fusaka/devnet2/crates/rpc-types-engine/src/payload.rs#L906
2. We need to send a `SignedBidSubmissionV5` to the relay overall: https://github.com/alloy-rs/alloy/blob/main/crates/rpc-types-beacon/src/relay.rs#L183

* [x] Run `make lint`
* [x] Run `make test`

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -49,51 +49,51 @@ codegen-units = 1
 incremental = false
 
 [workspace.dependencies]
-reth = { git = "https://github.com/bharath-123/reth", branch = "bharath/fusaka-revm-fix" }
-reth-chain-state = { git = "https://github.com/bharath-123/reth", branch = "bharath/fusaka-revm-fix" }
-reth-cli-util = { git = "https://github.com/bharath-123/reth", branch = "bharath/fusaka-revm-fix" }
-reth-db = { git = "https://github.com/bharath-123/reth", branch = "bharath/fusaka-revm-fix" }
-reth-db-common = { git = "https://github.com/bharath-123/reth", branch = "bharath/fusaka-revm-fix" }
-reth-errors = { git = "https://github.com/bharath-123/reth", branch = "bharath/fusaka-revm-fix" }
-reth-payload-builder = { git = "https://github.com/bharath-123/reth", branch = "bharath/fusaka-revm-fix" }
-reth-node-api = { git = "https://github.com/bharath-123/reth", branch = "bharath/fusaka-revm-fix" }
-reth-node-ethereum = { git = "https://github.com/bharath-123/reth", branch = "bharath/fusaka-revm-fix" }
-reth-trie = { git = "https://github.com/bharath-123/reth", branch = "bharath/fusaka-revm-fix" }
-reth-trie-parallel = { git = "https://github.com/bharath-123/reth", branch = "bharath/fusaka-revm-fix" }
-reth-basic-payload-builder = { git = "https://github.com/bharath-123/reth", branch = "bharath/fusaka-revm-fix" }
-reth-node-core = { git = "https://github.com/bharath-123/reth", branch = "bharath/fusaka-revm-fix" }
-reth-primitives = { git = "https://github.com/bharath-123/reth", branch = "bharath/fusaka-revm-fix" }
-reth-ethereum-primitives = { git = "https://github.com/bharath-123/reth", branch = "bharath/fusaka-revm-fix" }
-reth-primitives-traits = { git = "https://github.com/bharath-123/reth", branch = "bharath/fusaka-revm-fix" }
-reth-provider = { git = "https://github.com/bharath-123/reth", branch = "bharath/fusaka-revm-fix", features = [
+reth = { git = "https://github.com/paradigmxyz/reth", branch = "fusaka-devnet-3" }
+reth-chain-state = { git = "https://github.com/paradigmxyz/reth", branch = "fusaka-devnet-3" }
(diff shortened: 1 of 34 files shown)


Cargo.toml
```diff
@@ -49,94 +49,90 @@ codegen-units = 1
...
-reth-trie-db = { git = "https://github.com/paradigmxyz/reth", tag = "v1.4.8" }
-reth-payload-primitives = { git = "https://github.com/paradigmxyz/reth", tag = "v1.4.8" }
-reth-transaction-pool = { git = "https://github.com/paradigmxyz/reth", tag = "v1.4.8" }
-reth-execution-types = { git = "https://github.com/paradigmxyz/reth", tag = "v1.4.8" }
-reth-revm = { git = "https://github.com/paradigmxyz/reth", tag = "v1.4.8" }
-reth-payload-builder-primitives = { git = "https://github.com/paradigmxyz/reth", tag = "v1.4.8" }
-reth-payload-util = { git = "https://github.com/paradigmxyz/reth", tag = "v1.4.8" }
-reth-rpc-layer = { git = "https://github.com/paradigmxyz/reth", tag = "v1.4.8" }
-reth-testing-utils = { git = "https://github.com/paradigmxyz/reth", tag = "v1.4.8" }
+reth-chainspec = { git = "https://github.com/bharath-123/reth", branch = "bharath/fusaka-revm-fix" }
```
bharath-123: The current reth `fusaka-devnet-2` branch breaks with rbuilder because of breaking changes in revm with eip-7097. This should be resolved by `fusaka-devnet-3`. We are temporarily using a fork of reth for now.

crates/rbuilder/src/primitives/mod.rs
```diff
@@ -711,10 +713,26 @@ impl TransactionSignedEcRecoveredWithBlobs {
...
+                    metadata: metadata.unwrap_or_default(),
+                }),
+            }
         } else {
             Ok(Self {
                 tx,
-                blobs_sidecar: Arc::new(blob_sidecar.unwrap_or_default()),
+                blobs_sidecar: Arc::new(BlobTransactionSidecarVariant::Eip4844(
+                    BlobTransactionSidecar::default(),
+                )),
```
bharath-123, line 735: We ideally need to know which fork we are currently at, in order to create an EIP-4844 type sidecar or EIP-7954 sidecar. We unfortunately don't have this info here.

crates/rbuilder/src/primitives/mod.rs
```diff
@@ -742,17 +760,17 @@ impl TransactionSignedEcRecoveredWithBlobs {
         T: TransactionOrdering<Transaction = <V as TransactionValidator>::Transaction>,
         S: BlobStore,
     {
-        let blob_sidecar = pool
-            .get_blob(*tx.inner().hash())?
-            .and_then(|b| b.as_eip4844().cloned());
+        let mut blobs = pool.get_all_blobs(vec![*tx.inner().hash()])?;
```
bharath-123, line 763: `pool.get_blob` wasn't able to retrive the blob in my testing. using `pool.get_all_blobs` worked in this case.

ZanCorDX:
Covered by https://github.com/flashbots/rbuilder/pull/698

PR #671, opened 2025-07-29
feat: use at most builder balance for bid

efbig:
## 📝 Summary

Use at most builder balance for bid.

---

* [ ] Run `make lint`
* [ ] Run `make test`


dvush (changes requested):
I don't think that this will work because of delayed refunds.

I think its better to do this in a bidding service. 
Give it builder balance and make sure it never asks to bid more than a balance. 

Or more simply we can do that on our side where we receive response from a bidding service and cap payout to max of what we can pay.

This way we don't touch this brittle logic of payout out and we just modify the only input to it.

ZanCorDX:
Solved in the bidding service

PR #675, opened 2025-08-05
feat: sponsored bundles (#25)

quasarwin:
## 📝 Summary

This adds sponsorship of transactions that would otherwise fail with a LackOfFundForMaxFee.

We check the senders balance first to detect the lack of funds rather than wait on the revert and then have to resimulate.

## 💡 Motivation and Context

Transactions can fail because they lack funds, but they generate mev and would be actually to execute if they did have funds to do so.
---

* [x ] Run `make lint`
* [ x] Run `make test`
* [ x] Added tests (if applicable)

### DIFF
--- a/crates/rbuilder/src/building/order_commit.rs
+++ b/crates/rbuilder/src/building/order_commit.rs
@@ -35,6 +35,14 @@ use revm::{
 };
 use std::{collections::HashMap, sync::Arc};
 use thiserror::Error;
+use tracing::trace;
+
+/// Maximum amount that can be sponsored in a single transaction (2 ETH)
+const MAX_SPONSOR_AMOUNT: U256 = U256::from_limbs([2_000_000_000_000_000_000, 0, 0, 0]); // 2 * 10^18 wei = 2 ETH
+
+/// Safety buffer for sponsorship transactions to account for gas price changes,
+/// timing differences, and EVM precision issues (0.001 ETH)
+const SPONSORSHIP_SAFETY_BUFFER: U256 = U256::from_limbs([1_000_000_000_000_000, 0, 0, 0]); // 0.001 ETH
 
 #[derive(Clone)]
 pub struct BlockState {
@@ -322,6 +330,7 @@ pub struct PartialBlockFork<'a, 'b, 'c, 'd, Tracer: SimulationTracer> {
     tmp_used_state_tracer: UsedStateTrace,
 }
 
+#[derive(Clone)]
 pub struct PartialBlockRollobackPoint {
     rollobacks: usize,
 }
@@ -740,6 +749,7 @@ impl<'a, 'b, 'c, 'd, Tracer: SimulationTracer> PartialBlockFork<'a, 'b, 'c, 'd, 
         combined_refunds: &HashMap<Address, U256>,
     ) -> Result<Result<BundleOk, BundleErr>, CriticalCommitOrderError> {
         let mut refundable_profit = U256::ZERO;
+        let coinbase_balance_before = self.coinbase_balance()?;
         let mut insert = BundleOk {
             gas_used: 0,
             cumulative_gas_used,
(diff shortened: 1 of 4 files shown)


crates/rbuilder/src/building/order_commit.rs
```diff
@@ -754,6 +764,215 @@ impl<'a, 'b, 'c, 'd, Tracer: SimulationTracer> PartialBlockFork<'a, 'b, 'c, 'd,
...
+                            &mut self.local_ctx.cached_reads,
+                        )?;
+
+                        if builder_balance < total_sponsor_amount {
+                            // Skip sponsorship if builder balance is insufficient
+                            continue;
+                        }
+
+                        // Get the correct nonce accounting for previous sponsor transactions in this bundle
+                        let builder_nonce = {
```
dvush, line 818: this branch is not needed, its enough to just use else version of this branch

crates/rbuilder/src/building/order_commit.rs
```diff
@@ -754,6 +764,215 @@ impl<'a, 'b, 'c, 'd, Tracer: SimulationTracer> PartialBlockFork<'a, 'b, 'c, 'd,
...
+                            total_sponsor_amount,
+                        );
+
+                        if let Ok(unsigned_tx) = sponsor_tx_result {
+                            let sponsor_tx =
+                                TransactionSignedEcRecoveredWithBlobs::new_no_blobs(unsigned_tx)
+                                    .unwrap();
+
+                            // Validate nonce to prevent reuse
+                            let current_state_nonce = self.state.nonce(
```
dvush, line 852: why do you check nonce here, code above should give you correct nonce

crates/rbuilder/src/building/order_commit.rs
```diff
@@ -754,6 +764,215 @@ impl<'a, 'b, 'c, 'd, Tracer: SimulationTracer> PartialBlockFork<'a, 'b, 'c, 'd,
...
+
+                            // Execute sponsor tx.
+                            let sponsor_result = self.commit_tx(
+                                &sponsor_tx,
+                                insert.cumulative_gas_used,
+                                gas_reserved,
+                                insert.cumulative_blob_gas_used,
+                            )?;
+
+                            trace!(
```
dvush, line 871: all other logs in the builder use style like this

`trace!(?tx_hash, success = ..., "Sponsor tx execution result")`

crates/rbuilder/src/building/order_commit.rs
```diff
@@ -754,6 +764,215 @@ impl<'a, 'b, 'c, 'd, Tracer: SimulationTracer> PartialBlockFork<'a, 'b, 'c, 'd,
...
+                let shortfall = required_funds.saturating_sub(sender_balance);
+
+                // Add a fixed safety buffer to account for gas price variations and timing issues
+                let safety_buffer = SPONSORSHIP_SAFETY_BUFFER;
+                let total_sponsor_amount = shortfall + safety_buffer;
+
+                if shortfall > U256::ZERO {
+                    // Check if shortfall exceeds the maximum sponsor amount
+                    if total_sponsor_amount > MAX_SPONSOR_AMOUNT {
+                        continue;
```
dvush, line 797: If I read it correctly all these exits from the sponsorship branch like continue here are related to loop over txs in the bundle which means that this bundle tx will simply be skipped? If so then its not correct.

crates/rbuilder/src/building/order_commit.rs
```diff
@@ -754,6 +764,215 @@ impl<'a, 'b, 'c, 'd, Tracer: SimulationTracer> PartialBlockFork<'a, 'b, 'c, 'd,
...
+                                        // First the sponsor tx, then the user tx.
+                                        Self::accumulate_tx_execution(sponsor_ok, &mut insert);
+                                        // Extract values before moving res
+                                        let tx_success = res.tx_info.receipt.success;
+                                        let tx_coinbase_profit = res.tx_info.coinbase_profit;
+                                        Self::accumulate_tx_execution(res, &mut insert);
+
+                                        // Update refundable_profit based on ACTUAL transaction profits, not optimistic bundle profit
+                                        // Only add profit from the original transaction if it's refundable and successful
+                                        if bundle.is_tx_refundable(&tx_hash)
```
dvush, line 941: I wonder if all of these can be simplified dramatically. Currently it adds a new branch for committing sponsored txs while it should be possible to do it in the same way so things like coinbase profit and refundable profit are handled by the same code that its handled now. Otherwise its a bit hard to verify that its correct. 

For example, (see comment about `continue` above) I think this approach can lead to part of the bundle being committed while we should revert that bundle instead.

In other words, better approach would be so the sponsorhip path hits `let result = self.commit_tx(`

crates/rbuilder/src/building/order_commit.rs
quasarwin: ~~No thats not correct. We are prchecking if a tx needs sponsorship prior to simulation. So if it needs sponsorship, and qualifies for sponsorship (ie below our MAX_SPONSOR_AMOUNT and assuming builder balance can support it), it will then attempt to add a sponsorship tx prior to the transaction that would otherwise fail. When it hits continue here, we just skip trying to add a sponsorship tx, and importantly roll back the state.  So we're just skipping the shortfall loop, and that loop only attempts to add a sponsorship tx, it doesn't affect the tx that requires the sponsorship.~~

crates/rbuilder/src/building/order_commit.rs
quasarwin: ~~Assuming you mean the if statement on line 800... when coinbase_payment is set to true in the config, the coin bases signing capability will be removed (builder_signer will be None). The coinbase is needed to sign the prepayment transaction for the sponsorship.~~

crates/rbuilder/src/building/order_commit.rs
quasarwin: This is just defensive programming to ensure if there are any race conditions or alterations introduced accidentally in the future, we catch them. Our fork has a lot more trace logging I added at various points that would actually log this if it ever happened. If it did, we'd miss the slot and also cause demotions of optimistic relay submissions, so it's a critical error to catch.

crates/rbuilder/src/building/order_commit.rs
quasarwin: ~~So there should be no issues with part of the bundle being committed, see the above response. If a sponsorship tx added prior to an underfunded tx doesn't cause a successful tx (along with profit), then it rolls back to the point before the sponsorship attempt and proceeds normally, meaning the underfunded transaction will fail, causing the bundle to fail unless its in reverting/dropping hashes.~~

So I just realized I am looking at the code on our fork, not on this fork. This code is a fork from develop branch and I copied things over to this without double checking or testing it. Our fork has diverted significantly and the branching is different. So yes, I think as is this will cause a problem.   

The main goal with this approach was to prevent simulation overhead. If we wait until LackOfFundForGasLimit, then we'll have to create a sponsorship transactions before the tx, then retry the tx, and this can happen multiple times in a bundle. There are a lot of LackOfFundForGasLimit simulation errors and the vast majority of them are not profitable to sponsor, so when we tested it, it did significantly increase simulation load with almost no benefit. 

We also try not to make too many architecturally significant changes, that way future merges are easier for us. Because of that, we tend to prefer more code duplication rather than more cross cutting changes, as long as it doesn't affect the feature or the performance.  So if you guys feel like there's a better long term architecture here, that makes sense, and we'd probably merge that in and help support it moving forward.

crates/rbuilder/src/building/order_commit.rs
quasarwin: I just realized I misread this one.. You're referring to the builder nonce code branch. This is handling the case when we sponsor multiple transactions within a single bundle. So the flow should be:

1. Builder's onchain nonce is 100
2. First sponsored tx: uses nonce 100, updates nonces_updated to (builder, 100)
3. Second sponsored tx: finds (builder, 100) in nonces_updated, uses 101
4. Third sponsored tx: finds (builder, 101) in nonces_updated, uses 102

crates/rbuilder/src/building/order_commit.rs
quasarwin: See next comment, yes this is a bug introduced from extracting code from our fork to this one.

ZanCorDX (changes requested):
The main function is a huge piece of code with some continues which are error prone.
It feels that this could be implemented nicer with a function call what would handle the sponsoring and trying to have a single point for the tx execution instead of a copy/paste.

quasarwin:
> The main function is a huge piece of code with some continues which are error prone. It feels that this could be implemented nicer with a function call what would handle the sponsoring and trying to have a single point for the tx execution instead of a copy/paste.

100% agree. I think I only communicated this on the [issues thread](https://github.com/flashbots/rbuilder/issues/25#issuecomment-3070715676). This was in no way meant for you all to merge. Our fork is just too different (we forked almost a year ago). I just ported over some code to demonstrate how it's accomplished.  Thats also why I just left values hard coded, etc. 

Just want to demonstrate the gotchas on the implementation. The primary being checking the tx shortfall before simulation, and the nonce handling given multiple possible sponsorships and the fact the signer is usually randomized during simulation which doesn't work with sponsorship.

ZanCorDX:
> > The main function is a huge piece of code with some continues which are error prone. It feels that this could be implemented nicer with a function call what would handle the sponsoring and trying to have a single point for the tx execution instead of a copy/paste.
> 
> 100% agree. I think I only communicated this on the [issues thread](https://github.com/flashbots/rbuilder/issues/25#issuecomment-3070715676). This was in no way meant for you all to merge. Our fork is just too different (we forked almost a year ago). I just ported over some code to demonstrate how it's accomplished. Thats also why I just left values hard coded, etc.
> 
> Just want to demonstrate the gotchas on the implementation. The primary being checking the tx shortfall before simulation, and the nonce handling given multiple possible sponsorships and the fact the signer is usually randomized during simulation which doesn't work with sponsorship.

Thanks. I'll close it then and we'll come to it for reference when we implement it.

PR #682, opened 2025-08-15
chore: update reth to 1.6.0

cakevm:
## 📝 Summary

Updated reth latest stable v1.6.0 and fixed all compilation issues arising from the migration. I build it on top of my other PR #681.

- revm requires at least rust 1.88.0, so it is update to that at least. 1.89 would introduce more code changes, so i settled with that.
- Migrated to new Nibbles API (migration from SmallVec to U256-based representation) (In #666 it was more a backward compability, I tried to fix it, but happy if someone could run this with a node. I only have the bench and tests as indication that it works)
- Using arbitrary crate version 1.4.1 (in lock file) to resolve feature conflict for `alloy-primitives`, with 1.4.2 it seems to break
- Added `#[allow(clippy::large_enum_variant)]` attributes instead of boxing. Can not tell if it give a performance advantage but did not liked to make the PR to large by introduce box everywhere.
- Converted benchmark test data files from old nibbles array format to new `U256` format (did not included the migration script cause it is a one off task)
- Removed `imports_granularity` cause it is only available in nightly, and `make lint` would fail.
- Fixed all string format due to new rust version

## 💡 Motivation and Context

Stay current with the latest reth improvements and I decided to migrate the test files for `eth-sparse-mpt` to not have some custom deserialization. I took some inspiration for the migration from reth for handling the nibbles correct. Maybe there is some more room to tweak things with more inline, etc.

The nibbles format changed (when serialized to json):
```
before:
[7,1,9,10,14,9,13]

after:
{
  "length": 7,
  "nibbles": "0x719ae9d000000000000000000000000000000000000000000000000000000000"
}
```

There is also a cool talk about Nibbles here: https://youtu.be/LOxkZcASH0s?t=2447

---

* [x] Run `make lint`
* [x] Run `make test`


ZanCorDX:
Covered by https://github.com/flashbots/rbuilder/pull/698

PR #686, opened 2025-08-20
Added cfg time_to_keep_mempool_txs_secs + metrics

ZanCorDX:
## 📝 Summary

New parameter time_to_keep_mempool_txs_secs replaces old hardcoded constant BLOCKS_TO_KEEP_TXS.
New metric so measure the mem used by mempool txs (so we can tune time_to_keep_mempool_txs_secs).
Updated config docs with time_to_keep_mempool_txs_secs and some more.

## 💡 Motivation and Context

It was unethical to kill so young txs.

* [X] Run `make lint`
* [X] Run `make test`

### DIFF
--- a/crates/rbuilder/src/live_builder/base_config.rs
+++ b/crates/rbuilder/src/live_builder/base_config.rs
@@ -545,7 +545,7 @@ pub const DEFAULT_RETH_DB_PATH: &str = "/mnt/data/reth";
 /// This will update every 2.4 hours, super reasonable.
 pub const DEFAULT_BLOCKLIST_URL_MAX_AGE_HOURS: u64 = 24;
 pub const DEFAULT_REQUIRE_NON_EMPTY_BLOCKLIST: bool = false;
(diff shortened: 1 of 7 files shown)


crates/rbuilder/src/live_builder/base_config.rs
```diff
@@ -542,6 +545,7 @@ pub const DEFAULT_RETH_DB_PATH: &str = "/mnt/data/reth";
 /// This will update every 2.4 hours, super reasonable.
 pub const DEFAULT_BLOCKLIST_URL_MAX_AGE_HOURS: u64 = 24;
 pub const DEFAULT_REQUIRE_NON_EMPTY_BLOCKLIST: bool = false;
+pub const DEFAULT_TIME_TO_KEEP_MEMPOOL_TXS_SECS: u64 = 1;
```
dvush: Why default is 1 sec? We kept mempool txs for couple of blocks before

ZanCorDX: Je, I used the wrong unit! It should be 1 min. Thanks!

dvush (approved):

PR #704, opened 2025-09-08
Streamline bidding interface

dvush:
## 📝 Summary

This PR replaces interfaces related to unfinished blocks processing and bidding service with more concrete implementations. 

All logic is consolidated in one module unfinished_block_processing.

As a result logic for sealing blocks and interactive with bidding service is simplified and becomes easier to track. For example, it moves logic for handling unfinished blocks from rbuilder-operator to this repo otherwise to track path of the block you would need to jump through traits that are implemented in two different repos (while we had only one real useful implementation of the flow). 

This PR removes 600 loc from this project and around 500 from rbuilder-operator without change of functionality. It also fixed one small but subtle issue with block sealing that probably caused us to seal blocks that are stale. 

This is requirement for making a PR for faster bidding / sealing. This PR does not change any functionality is a significant way. 

## 💡 Motivation and Context

Simplify flow of blocks to improve them in the next PR.

---
## TODO

* [ ] change block flow docs

* [ ] Run `make lint`
* [ ] Run `make test`


ZanCorDX (approved):

slerfonbase (approved):

dvush:
@ZanCorDX I'll address your comments in the next PR that I create this week. I'll merge the change for now.

PR #721, opened 2025-09-24
feat: delayed refunds

efbig:
## 📝 Summary

Add a new bundle field which allows orderflow senders to opt into being refunded via BuilderNet refund pipeline thus saving the gas costs on the payout transaction.

* [x] Run `make lint`
* [x] Run `make test`

### DIFF
--- a/crates/rbuilder/src/building/mod.rs
+++ b/crates/rbuilder/src/building/mod.rs
@@ -539,8 +539,10 @@ pub struct PartialBlock<
     pub coinbase_profit: U256,
     /// Tx execution info belonging to successfully executed orders.
     pub executed_tx_infos: Vec<TransactionExecutionInfo>,
-    /// Combined refunds.
+    /// Combined refunds to be paid at the end of the block.
     pub combined_refunds: HashMap<Address, U256>,
+    /// Cumulative delayed refund value which will be refunded via BuilderNet refund pipeline.
+    pub delayed_refund: U256,
     pub tracer: Tracer,
     partial_block_execution_tracer: PartialBlockExecutionTracerType,
 }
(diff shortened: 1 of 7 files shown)


crates/rbuilder/src/building/mod.rs
```diff
@@ -539,8 +539,10 @@ pub struct PartialBlock<
     pub coinbase_profit: U256,
     /// Tx execution info belonging to successfully executed orders.
     pub executed_tx_infos: Vec<TransactionExecutionInfo>,
-    /// Combined refunds.
+    /// Combined refunds to be paid at the end of the block.
     pub combined_refunds: HashMap<Address, U256>,
+    /// Cumulative delayed refund value which will be refunded via BuilderNet refund pipeline.
+    pub delayed_refund: U256,
```
dvush, line 545: why is this needed?

crates/rbuilder/src/building/testing/bundle_tests/mod.rs
```diff
@@ -402,6 +403,83 @@ fn test_bundle_combined_refunds() -> eyre::Result<()> {
...
+            setup.set_bundle_refund(BundleRefund {
+                recipient,
+                percent: refund_percent,
+                tx_hash: profit_tx_hash,
+                delayed: true,
+            });
+            Ok(setup.commit_order_ok())
+        };
+
+    let result = commit_refund_order(&mut test_setup, recipient_address).unwrap();
```
dvush, line 434: why is this not a loop

dvush (changes requested):
This info needs to available in BuiltBlockTrace, so it should be in the ExecutionResult for the corresponding order

dvush (approved):

PR #726, opened 2025-09-26
feat: rebalancer service

efbig:
## 📝 Summary

Create a standalone rebalancer service for moving funds around BuilderNet accounts. After https://github.com/flashbots/rbuilder/pull/738 rbuilder would correctly be able to process rebalancing transactions from builder EOA without adding the value to the coinbase profit (as long as rebalancing targets are allowlisted).

* [x] Run `make lint`
* [x] Run `make test`

### DIFF
--- a/crates/rbuilder-rebalancer/src/rebalancer.rs
+++ b/crates/rbuilder-rebalancer/src/rebalancer.rs
@@ -1,0 +1,279 @@
+use alloy_consensus::{
+    EthereumTxEnvelope, EthereumTypedTransaction, SignableTransaction as _, TxEip1559, TxEip4844,
+};
+use alloy_eips::{eip1559::BaseFeeParams, BlockId, Encodable2718};
+use alloy_primitives::{
+    hex,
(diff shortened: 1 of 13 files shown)


crates/rbuilder-rebalancer/src/rebalancer.rs
```diff
@@ -0,0 +1,279 @@
...
+            accounts,
+            rules,
+            timeout,
+            retry_delay,
+            tracked_accounts: OnceCell::new(),
+        }
+    }
+
+    fn tracked_accounts(&self) -> &AddressSet {
+        self.tracked_accounts.get_or_init(|| {
```
dvush, line 60: why do you use once cell here?

dvush (approved):

crates/rbuilder-rebalancer/src/rebalancer.rs
efbig: because the tracked accounts are immutable

PR #751, opened 2025-10-08
fix: avoid filtering out system orders

efbig:
## 📝 Summary

Avoid filtering out system orders due to prio fee difference. This would help with system order inclusion. 

* [x] Run `make lint`
* [x] Run `make test`

### DIFF
--- a/crates/rbuilder-primitives/src/lib.rs
+++ b/crates/rbuilder-primitives/src/lib.rs
@@ -40,32 +40,66 @@ pub use test_data_generator::TestDataGenerator;
 use thiserror::Error;
 use uuid::Uuid;
 
-/// Extra metadata for ShareBundle/Bundle.
+/// Extra metadata for an order.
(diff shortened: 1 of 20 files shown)


crates/rbuilder-primitives/src/lib.rs
```diff
@@ -1226,6 +1226,8 @@ impl SimValue {
 #[derive(Debug, Clone, PartialEq, Eq)]
 pub struct SimulatedOrder {
     pub order: Order,
+    /// Flag indicating whether it's a system order.
+    pub is_system: bool,
```
dvush: I think its better to add this to order metadata (to its property of the order), it feels a bit out of place in simulated order.  Ideally we remove system_recipient_allowlist from ctx and populate that field on the rpc level so order is already prepared.

ZanCorDX (approved):

PR #760, opened 2025-10-10
feat(primitives): decoding with signer lookups

thedevbirb:
## 📝 Summary

Modifies the `rbuilder-primitives` crate so that it is possible to decode bundles specifying a signer lookup, to save on signature recovery compute.

## 💡 Motivation and Context

Most bundles received contain transactions already heard, as such we can reduce a lot of compute from the hot path.

**Summary of the changes**:
- `RawBundle` now feature a `decode_with_signer_lookup` that allows to specify a closure to check whether the signer is already known for a certain hash.
	- It calls an underlying `RawBundle::decode_inner` that takes this the closure as an `Option`.
- Inside `RawBundle::decode_inner`, an helper struct `RawTransactionDecodable` (naming can be improved) that contains the raw bytes, the `TxEncoding` and the optional signer lookup is used for decoding. Transaction decoding and recovering can't really be decoupled from bundle decoding because it's required for parsing refunds data.
- A `RawBundleMetadata` struct has been introduced to simplify a bit the code and for separation of concerns.

---

* [x] Run `make lint`
* [x] Run `make test`

### DIFF
--- a/crates/rbuilder-primitives/src/lib.rs
+++ b/crates/rbuilder-primitives/src/lib.rs
@@ -912,6 +912,49 @@ impl TransactionSignedEcRecoveredWithBlobs {
             }
         }
     }
+
+    /// Decodes the "raw" format of transaction (e.g. `eth_sendRawTransaction`) with the blob data (network format)
+    pub fn decode_enveloped_with_real_blobs_with_lookup(
+        raw_tx: Bytes,
+        lookup: impl Fn(B256) -> Option<Address>,
+    ) -> Result<TransactionSignedEcRecoveredWithBlobs, TxWithBlobsCreateError> {
+        let raw_tx = &mut raw_tx.as_ref();
+        let pooled_tx = PooledTransactionVariant::decode_2718(raw_tx)
+            .map_err(TxWithBlobsCreateError::FailedToDecodeTransaction)?;
+
+        let signer = if let Some(signer) = lookup(*pooled_tx.tx_hash()) {
+            signer
+        } else {
+            pooled_tx
+                .recover_signer()
+                .map_err(|_| TxWithBlobsCreateError::InvalidTransactionSignature)?
+        };
+
+        match pooled_tx {
+            PooledTransactionVariant::Legacy(_)
+            | PooledTransactionVariant::Eip2930(_)
+            | PooledTransactionVariant::Eip1559(_)
+            | PooledTransactionVariant::Eip7702(_) => {
+                let tx_signed = TransactionSigned::from(pooled_tx);
+                TransactionSignedEcRecoveredWithBlobs::new_no_blobs(tx_signed.with_signer(signer))
+            }
+            PooledTransactionVariant::Eip4844(blob_tx) => {
+                let (blob_tx, signature, hash) = blob_tx.into_parts();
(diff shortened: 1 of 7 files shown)


crates/rbuilder-primitives/src/lib.rs
```diff
@@ -877,59 +931,115 @@ impl TransactionSignedEcRecoveredWithBlobs {
...
+        let signer = if let Some(ref lookup) = self.signer_lookup {
+            if let Some(signer) = lookup.lookup(*pooled_tx.tx_hash()) {
+                signer
+            } else {
+                recover_signer()?
             }
-        }
+        } else {
+            recover_signer()?
+        };
```
mempirate:
```suggestion
        let signer = self
            .signer_lookup
            .as_ref()
            .and_then(|lookup| lookup.lookup(*pooled_tx.tx_hash()))
            .or_else(|| pooled_tx.recover_signer().ok())
            .ok_or(TxWithBlobsCreateError::InvalidTransactionSignature)?;
```

Another slightly shorter option

crates/rbuilder-primitives/src/lib.rs
```diff
@@ -877,59 +931,115 @@ impl TransactionSignedEcRecoveredWithBlobs {
...
-                .push(Bytes48::from([0u8; BYTES_PER_COMMITMENT]));
-            fake_sidecar
-                .proofs
-                .push(Bytes48::from([0u8; BYTES_PER_PROOF]));
-        }
+
+        let hash = *decoded.hash();
+
+        // TODO: clean this after rust 1.89 and if let chains
+        let tx = if let Some(ref lookup) = self.signer_lookup {
```
mempirate: Same as above

crates/rbuilder-primitives/src/lib.rs
```diff
@@ -877,59 +931,115 @@ impl TransactionSignedEcRecoveredWithBlobs {
         self.tx.encode_2718(&mut buf);
         buf.into()
     }
+}
+
+pub trait SignerLookup {
```
mempirate: Curious, why a trait as opposed to a type alias to `Fn(B256) -> Option<Address>`?

mempirate (commented):
Initial pass

crates/rbuilder-primitives/src/lib.rs
thedevbirb: `Fn(B256) -> Option<Address>` is trait, but in practice I managed to simplify it a little bit by having the `NoOpSignerLookup`.

thedevbirb: [daddfa2](https://github.com/flashbots/rbuilder/pull/760/commits/daddfa209c1c944a1c08fcb2b21c7ce98926462f)

crates/rbuilder-primitives/src/lib.rs
thedevbirb: [45f9827](https://github.com/flashbots/rbuilder/pull/760/commits/45f98276334057aaa190ffdf839fa8f5717e61a5)

efbig (approved):
lgtm

dvush (approved):

mempirate (approved):
lgtm

PR #764, opened 2025-10-13
Trace warn if order execution  takes to long

ZanCorDX:
## 📝 Summary

This PR add a new config:
max_order_execution_duration_warning_us (optional u64): If set, while building a block if an order execution duration is greater than this, we will log a warning with some info about the order.

* [X] Run `make lint`
* [X] Run `make test`

### DIFF
--- a/crates/rbuilder/src/building/builders/block_building_helper.rs
+++ b/crates/rbuilder/src/building/builders/block_building_helper.rs
@@ -1,4 +1,4 @@
-use alloy_primitives::{utils::format_ether, U256};
+use alloy_primitives::{utils::format_ether, Address, TxHash, U256};
 use reth_provider::StateProvider;
 use std::{
     cmp::max,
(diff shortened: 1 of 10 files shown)


crates/rbuilder/src/building/builders/block_building_helper.rs
```diff
@@ -504,6 +512,12 @@ impl<
             Err(e) => (Err(e), false),
         };
         add_order_simulation_time(sim_time, &self.builder_name, sim_ok);
+        if self
+            .max_order_execution_duration_warning
+            .is_some_and(|max_dur| sim_time > max_dur)
+        {
+            warn!(sim_ok,?sim_time,builder_name=self.builder_name,id = ?order.id(),txs = ?order.order.list_txs().iter().map(|(tx, _)| tx.hash()).collect::<Vec<_>>(), "Slow order execution");
```
dvush: I found the following fields to be useful:
* tx from / to (can be only from the first tx or from all txs like you do here)
*bundle signer

dvush (approved):

crates/rbuilder/src/building/builders/block_building_helper.rs
reaper-o-lykos: Great idea!!

PR #780, opened 2025-10-23
Clickhouse backup on utils

ZanCorDX:
## 📝 Summary

Brought clickhouse robust saving (with local copy) from orderflow proxy repo (flashbots/buildernet-orderflow-proxy-v2.git)

## 💡 Motivation and Context

Code reuse

---

* [X] Run `make lint`
* [X] Run `make test`


crates/rbuilder-utils/src/clickhouse/backup/mod.rs
```diff
@@ -0,0 +1,737 @@
...
+    /// The interval at which we try to backup data.
+    interval: BackoffInterval,
+
+    /// A failed commit retrieved from either disk or memory, waiting to be retried.
+    last_cached: Option<RetrievedFailedCommit<T>>,
+
+    /// Whether to use only the in-memory backup (for testing purposes).
+    #[cfg(test)]
+    use_only_memory_backup: bool,
+    _metrics_phantom: std::marker::PhantomData<MetricsType>,
```
thedevbirb, line 525: Interesting, how are you planning to use this? I mean the `Metrics` trait and the marker here.

crates/rbuilder-utils/src/backoff.rs
thedevbirb, line 1: Why is this inside a `metrics` folder? This is more `time`-related utilities.

crates/rbuilder-utils/src/clickhouse/indexer.rs
thedevbirb, line 1: What's the rationale for distinguish `clickhouse` and `clickhouse_with_backup`? Eventually, I get it could make sense to have a `backup` subfolder of `clickhouse`.

crates/rbuilder-utils/src/tokio/mod.rs
thedevbirb, line 1: In orderflow proxy v2, this has been extracted from `reth_tasks` so that the project didn't have reth dependencies (at least directly). However, rbuilder heavily depends on reth already. As such, I'd suggest you to import `reth_tasks` in this project and use that instead. Then, you can delete this folder entirely.

thedevbirb (commented):
First pass!

crates/rbuilder-utils/src/clickhouse_with_backup/metrics.rs
```diff
@@ -0,0 +1,17 @@
...
+    fn record_batch_commit_time(duration: Duration);
+    fn increment_commit_failures(err: String);
+    fn set_queue_size(size: usize, order: &'static str);
+    fn set_disk_backup_size(size_bytes: u64, batches: usize, order: &'static str);
+    fn increment_backup_disk_errors(order: &'static str, error: &str);
+    fn set_memory_backup_size(size_bytes: u64, batches: usize, order: &'static str);
+    fn process_backup_data_lost_quantities(quantities: &Quantities);
+    fn process_backup_data_quantities(quantities: &Quantities);
+    fn set_backup_empty_size(order: &'static str);
+}
```
thedevbirb: You can expose a `NoOp` struct here with default implementation. Also another nit/question, do you explicitly want to not provide a default, empty implementation?

crates/rbuilder-utils/src/tokio/mod.rs
thedevbirb: In more detail, rbuilder should re-export `reth_tasks` so that orderflow proxy v2 can use it, so there is no need to need to add it again as a dependency.

thedevbirb (commented):
Second pass

thedevbirb (approved):
On a second thought, maybe it makes more sense to move `macros.rs`, `primitives.rs` and `metrics.rs` up to `src/clickhouse`, and not inside `backup`. Rationale being it not necessarily backup-related logic.

In any case, this is okay, and I don't care about nitpicking too much right now as this will be only used by of-proxy for the time being.

rkrasiuk (approved):

efbig (approved):

PR #785, opened 2025-10-27
feat: add ace protocols

Will-Smith11:
## 📝 Summary

Adjusted the ingestion pipeline to allow for specific interactions with known ace protocols to modify the simulation process to allow orders, in which we know are valid with a defined ace protocol through. Also some minor changes to the block building process putting protocol ace tx's that don't interact with state at the top

## 💡 Motivation and Context
Currently Other builders have started to include ace protocols and we want to also have buildernet cover this. Ace protocols have specific sequencing rules enforced on-chain that allow them to dictate execution order. By allowing for ace protocols to integrate into builders, It unlocks more transactions to execute in the same block that wouldn't be able to otherwise. The increase in transaction flow should help the builder stay competitive.
---

## Questions
This is my first time in the codebase. I wanted to open up this PR so I could get some feedback to ensure that with the given architecture that everything is implemented in the right locations. I will add a bunch of tests once I know that the architecture is solid, just don't want to do a bunch more work when I'm uncertain about design.

* [x] Run `make lint`
* [x] Run `make test`


crates/rbuilder/src/building/sim.rs
```diff
@@ -299,14 +531,163 @@ impl SimTree {
         Ok(())
     }
 
+    /// Handle ACE unlocking interaction after successful simulation.
+    /// Returns optional cancellation OrderId if a mempool unlock cancels an optional ACE tx.
+    /// Note: NonUnlocking ACE interactions are handled at the OrderSimResult level.
+    pub fn handle_ace_unlock(
```
ZanCorDX, line 742: Should take as args simulated_order,previous_orders since it's always called in match SimulatedResult::Success no need to recheck that.

crates/rbuilder/src/building/sim.rs
```diff
@@ -299,14 +531,163 @@ impl SimTree {
...
+        &mut self,
+        result: &SimulatedResult,
+    ) -> Result<Vec<OrderId>, ProviderError> {
```
ZanCorDX, line 745: Never returns error!!! Remove Result

crates/rbuilder/src/building/sim.rs
```diff
@@ -299,14 +531,163 @@ impl SimTree {
...
+                AceUnlockSource::ProtocolOptional => {
+                    let state = self.ace_state.entry(contract_address).or_default();
+
+                    // Check if user unlock already available - cancel optional
+                    if state.has_mempool_unlock {
+                        trace!(
+                            "Cancelling optional ACE unlock for {:?} - user unlock exists",
+                            contract_address
+                        );
+                        cancellations.push(simulated_order.order.id());
```
ZanCorDX, line 799: It's cancelling it's own order?
It should only cancell the order returned by mark_mempool_unlock.
If it wants to express that the order analized should not be inserted it should return something like a bool saying it so we don’t send and order and it’s cancellation.

crates/rbuilder/src/building/sim.rs
```diff
@@ -299,14 +531,163 @@ impl SimTree {
...
+            simulated_order,
+            previous_orders,
+            ..
+        } = result
+        else {
+            return Ok(Vec::new());
+        };
+
+        // If this order already has parents, it was re-simulated - just pass through
+        if !previous_orders.is_empty() {
```
ZanCorDX, line 756: Will return for a user tx having a previous nonce dep, are we ok with this?

crates/rbuilder/src/building/sim.rs
```diff
@@ -299,14 +531,163 @@ impl SimTree {
...
+                    }
+
+                    // Store optional unlock - there are orders waiting for it
+                    state.optional_unlock_order = Some(simulated_order.clone());
+                    trace!(
+                        "Added optional ACE protocol unlock order for {:?}",
+                        contract_address
+                    );
+                }
+                AceUnlockSource::User => {
```
ZanCorDX, line 821: There is NO guaranty that a AceUnlockSource::User will be executed before orders depending o it.
A depending order could make a lot of money so it could be placed before the AceUnlockSource::User.

crates/rbuilder/src/building/sim.rs
```diff
@@ -299,14 +531,163 @@ impl SimTree {
...
+
+    /// Process simulation results, handling ACE unlocks and updating dependencies.
+    /// Returns:
+    /// - `Vec<SimulatedResult>`: Successful results (to be forwarded to builder)
+    /// - `Vec<OrderId>`: Order IDs that should be cancelled (e.g., optional ACE unlocks superseded by mempool)
     pub fn submit_simulation_tasks_results(
         &mut self,
         results: Vec<SimulatedResult>,
-    ) -> Result<(), ProviderError> {
+    ) -> Result<(Vec<SimulatedResult>, Vec<OrderId>), ProviderError> {
```
ZanCorDX, line 859: nobody uses the Vec<SimulatedResult> in the result. Remove it.

crates/rbuilder/src/building/sim.rs
```diff
@@ -369,36 +751,61 @@ where
...
-                        ?err,
-                        "Order simulation failed"
-                    );
-                    sim_errors.push(err);
-                    continue;
+                OrderSimResult::Failed(failure) => {
+                    if let Some(contract_address) = failure.ace_dependency {
+                        // Order failed but needs ACE unlock - queue for re-simulation
+                        // Pass existing ace_unlock_contracts for progressive multi-ACE discovery
+                        sim_tree.add_ace_dependency_for_order(
```
ZanCorDX: Why simulate_all_orders_with_sim_tree is calling "sim_tree.add_ace_dependency_for_order" manually? Isn't submit_simulation_tasks_results doing that already?

crates/rbuilder/src/building/sim.rs
```diff
@@ -276,21 +500,29 @@ impl SimTree {
             }
         }
 
-        for ready_order in orders_ready {
-            let pending_state = self.get_order_nonce_state(&ready_order)?;
+        for mut ready_pending_order in orders_ready {
+            let pending_state = self.get_order_dependency_state(&ready_pending_order.order)?;
             match pending_state {
-                OrderNonceState::Ready(parents) => {
+                OrderDependencyState::Ready(mut parents) => {
```
ZanCorDX, line 703: Here parents contains only Nonce dependencies and not any other prev ACE dependency.
If the order was depending in another ACE dependency that will not be included.
If looks like you should add all deps in PendingOrder::ace_unlock_contracts not only this one last unlocked.

crates/rbuilder/src/building/sim.rs
```diff
@@ -47,28 +64,101 @@ pub struct NonceKey {
...
+        }
+    }
+}
+
 #[derive(Debug, Clone, PartialEq, Eq)]
 struct PendingOrder {
     order: Order,
-    unsatisfied_nonces: usize,
+    unsatisfied_dependencies: usize,
+    /// ACE contracts already provided as unlock parents (for progressive multi-ACE discovery)
```
ZanCorDX: Shouldn't this commet just be "ACE contract needed to be unlocked"?
You can have this pending without alread having an unlock parent.....

crates/rbuilder/src/building/sim.rs
```diff
@@ -187,27 +313,96 @@ impl SimTree {
...
+        mut existing_ace_unlock_contracts: HashSet<Address>,
+    ) -> Result<(), ProviderError> {
+        // Add new contract to the set
+        existing_ace_unlock_contracts.insert(new_contract);
+        let dep_key = DependencyKey::AceUnlock(new_contract);
+
+        // Check if we already have an unlock provider for the new contract
+        if self.dependency_providers.contains_key(&dep_key) {
+            // Build parents from ALL ACE unlock contracts we need
+            let mut parents = Vec::new();
```
ZanCorDX: You are not adding Nonce deps here.

ZanCorDX:
From the comment I made in add_ace_dependency_for_order and process_simulation_task_result it feels that you need a common func that computes the parent.
Maybe get_order_dependency_state gets renamed to get_order_nonce_dependency_state (since it only computes nonce deps and the new function get_order_dependency_state takes an extra parameter with the ACE needed contracts and also calls get_order_dependency_state ?

crates/rbuilder/src/building/sim.rs
```diff
@@ -453,32 +872,138 @@ pub fn simulate_order_using_fork<Tracer: SimulationTracer>(
...
+
+    // Detect ACE interactions from the state trace using config
+    // Check ALL transactions in the order and collect ALL ACE interactions (one per contract)
+    // For each contract, keep the highest priority classification:
+    // Priority: ProtocolForce > ProtocolOptional > User > NonUnlocking
+    let ace_interactions: Vec<AceInteraction> = if let Some(trace) = used_state_trace.as_ref() {
+        // Use HashMap to track best interaction per contract
+        let mut per_contract: HashMap<Address, AceInteraction> = HashMap::default();
+
+        for (tx, _) in order.list_txs() {
```
ZanCorDX, line 1090: All txts are being iterated, even if they where droped

crates/rbuilder/src/building/sim.rs
```diff
@@ -453,32 +872,138 @@ pub fn simulate_order_using_fork<Tracer: SimulationTracer>(
...
+            } else {
+                None
+            };
+            let tx_to = tx.to();
+            let tx_from = Some(tx.signer());
+
+            // Check this transaction against all ACE configs
+            for (_, config) in ace_configs.iter() {
+                if let Some(interaction) =
+                    classify_ace_interaction(trace, sim_success, config, selector, tx_to, tx_from)
```
ZanCorDX, line 1103: sim_success is for the whole Order, it's not related to the particular execution of this Tx.
This might be a problem since we don't have at the moment trace state per child tx.
Eg: For a Bundle we could have a Dropeable/Revertable Tx accessing the pool and failing but the bundle succeding so we would have classify_ace_interaction answering Unlocking.
To detect Force/Optional unlocking orders we could asume that Order must be a single Tx.

crates/rbuilder/src/building/sim.rs
Will-Smith11: This should be fine. otherwise we are making a lot more downstream changes

PR #806, opened 2025-11-07
Improved reproducible builds

bakhtin:
## 📝 Summary
- Switch release pipeline to building inside the Docker container for better reproducibility. Previous implementation produced reproducible builds only on the same host. Building inside the Docker container allows producing the same binary on different hosts and even different OSes.
- Set Docket image Rust version for building the binary by parsing a version configured for the project. That is to avoid manually updating Docker images when the Rust version bumps.
- Add reproducible testing CI job that runs once every 2 days building on two different hosts and comparing the hashsums of produced binaries.
- Switch to using jemalloc-sys from Debian repos instead of building it from source. A Debian version is [reproducible](https://tests.reproducible-builds.org/debian/rb-pkg/trixie/amd64/jemalloc.html) which is [hard to achieve](https://github.com/NixOS/nixpkgs/issues/380852) if you build it from source. It does not require a runtime dependency as the library links statically.

## 💡 Motivation and Context

---

* [ ] Run `make lint`
* [ ] Run `make test`

### DIFF
--- a/.github/workflows/reprotest.yml
+++ b/.github/workflows/reprotest.yml
@@ -1,0 +1,78 @@
+name: reproducible-build-test
+
+on:
+  workflow_dispatch: {}
+  schedule:
+    - cron: "0 1 */2 * *"
+
+jobs:
+  build:
+    name: build reproducible binaries
+    runs-on: ${{ matrix.runner }}
+    strategy:
+      matrix:
+        include:
+          - runner: warp-ubuntu-2404-x64-32x
+            machine: machine-1
+          - runner: warp-ubuntu-2204-x64-32x
+            machine: machine-2
+    steps:
+      - uses: actions/checkout@v5
+
+      - name: Set up Docker Buildx
+        uses: docker/setup-buildx-action@v3
+
+      - name: Install rust
+        run: |
+          curl --proto '=https' --tlsv1.2 -sSf https://sh.rustup.rs | sh -s -- -y
+
+      - name: Build reproducible binary with Docker
+        run: |
+          RUST_TOOLCHAIN=$(rustc --version | cut -d' ' -f2)
+          docker build \
+            --build-arg "RUST_TOOLCHAIN=${RUST_TOOLCHAIN}" \
+            -f docker/Dockerfile.reproducible -t rbuilder:release \
+            --output type=local,dest=./target .
+
+      - name: Calculate SHA256
+        id: sha256
+        run: |
+          sha256sum target/reproducible/{rbuilder-operator,rbuilder-rebalancer} > checksums.sha256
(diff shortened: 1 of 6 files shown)


rust-toolchain.toml
```diff
@@ -1,4 +1,3 @@
 [toolchain]
 channel = "stable"
-version = "1.88.0"
```
bakhtin, line 3: `version` is not recognized as the valid option and ignored: https://rust-lang.github.io/rustup/overrides.html#the-toolchain-file

.github/workflows/reprotest.yml
```diff
@@ -0,0 +1,78 @@
...
+      - name: Set up Docker Buildx
+        uses: docker/setup-buildx-action@v3
+
+      - name: Install rust
+        run: |
+          curl --proto '=https' --tlsv1.2 -sSf https://sh.rustup.rs | sh -s -- -y
+
+      - name: Build reproducible binary with Docker
+        run: |
+          RUST_TOOLCHAIN=$(rustc --version | cut -d' ' -f2)
```
ilyaluk, line 31: If rust is installed only for latest version, would it be better to just pin the version here instead of installing toolchain?

Same for release.yaml.

ZanCorDX (approved):

docker/Dockerfile.reproducible
```diff
@@ -0,0 +1,18 @@
...
+RUN sed -i '/^# http/{N;s|^# \(http[^ ]*\)\nURIs: .*|# \1\nURIs: \1|}' /etc/apt/sources.list.d/debian.sources
+RUN apt-get -o Acquire::Check-Valid-Until=false update && \
+    apt-get install -y \
+        libjemalloc-dev \
+        libclang-dev \
+        protobuf-compiler \
+        cmake
+WORKDIR /build
+COPY . .
+RUN SOURCE_DATE=1730000000 make build && make build-deb
```
ilyaluk, line 15: Maybe use last commit timestamp as SOURCE_DATE?

.github/workflows/reprotest.yml
bakhtin: I install rust toolchain just to parse rust-toolchain.yml file and pull the correct Docker/Rust image. rust-toolchain may contain something like `stable` or a numerical version of Rust. So it is hard to directly parse the file without (some) rust toolchain installed

docker/Dockerfile.reproducible
bakhtin: It may not even be a git tree (e.g., if downloaded as a tar archive from GH) so I don't want to rely on it

ilyaluk (approved):

PR #809, opened 2025-11-11
Relay list disabled on blockprocessor calls

ZanCorDX:
## 📝 Summary

This allows us to deploy with the Multibid features disabled without updating blockprocessor.

### DIFF
--- a/crates/rbuilder-operator/src/blocks_processor.rs
+++ b/crates/rbuilder-operator/src/blocks_processor.rs
@@ -93,8 +93,8 @@ type ConsumeBuiltBlockRequest = (
     // bestBidValue
     U256,
     Vec<BlockProcessorDelayedPayments>,
-    // Relays
-    Vec<String>,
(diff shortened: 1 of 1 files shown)


crates/rbuilder-operator/src/blocks_processor.rs
```diff
@@ -93,8 +93,8 @@ type ConsumeBuiltBlockRequest = (
     // bestBidValue
     U256,
     Vec<BlockProcessorDelayedPayments>,
-    // Relays
-    Vec<String>,
```
dvush, line 97: please don't remove, this might break API that receives it

dvush (changes requested):

crates/rbuilder-operator/src/blocks_processor.rs
ZanCorDX: The API was never updated so I'm reverting the change until we do.

dvush (approved):

PR #835, opened 2025-12-16
feat: allow OTLP exporting by extending existing logger configs

prestwich:
## 📝 Summary

Add simple OTLP exporting by extending existing logger configs.

The goal is to allow rbuilder to export spans, events, and other trace information to tools like zipkin or groundcover

## 💡 Motivation and Context

Distributed tracing is important for production systems :)

* [x] Run `make lint`
* [x] Run `make test`

### DIFF
--- a/crates/rbuilder-config/src/logger.rs
+++ b/crates/rbuilder-config/src/logger.rs
@@ -1,4 +1,11 @@
-use tracing_subscriber::EnvFilter;
+use tracing_subscriber::{
+    fmt::{
+        format::{DefaultFields, Format, Full, Json, JsonFields},
+        time::SystemTime,
+        SubscriberBuilder,
+    },
+    EnvFilter,
+};
 
 /// Logger configuration.
 #[derive(PartialEq, Eq, Clone, Debug, serde::Deserialize)]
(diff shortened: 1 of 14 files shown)


crates/rbuilder-config/src/logger.rs
```diff
@@ -22,15 +29,57 @@ impl LoggerConfig {
...
+    }
+
+    /// Create an ANSI-formatting [`SubscriberBuilder`] from the configuration.
+    ///
+    /// Errors if the filter string is invalid.
+    ///
+    /// # Panics
+    ///
+    /// If `self.log_json` is true.
+    pub(crate) fn ansi(&self) -> eyre::Result<SubscriberBuilder<DefaultFields, Format, EnvFilter>> {
```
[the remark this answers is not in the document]
prestwich: i go the other way. including default type args you don't have to is overly verbose

crates/rbuilder-config/src/tracing.rs
```diff
@@ -0,0 +1,128 @@
...
+
+/// Install a format layer based on the logger configuration, and then install
+/// the registry.
+macro_rules! add_fmt_and_install {
+    (json @ $registry:ident, $filter:ident) => {{
+        let fmt = tracing_subscriber::fmt::layer().json().with_filter($filter);
+        $registry.with(fmt).try_init()
+    }};
+    (log @ $registry:ident, $filter:ident) => {{
+        let fmt = tracing_subscriber::fmt::layer().with_filter($filter);
```
ZanCorDX: did you forget the ansi() ?

prestwich: ya, it needs to be propagated through.

incidentally, the preferred config method for ANSI is [`NO_COLOR`](https://no-color.org/). As documented [here](https://docs.rs/tracing-subscriber/latest/tracing_subscriber/fmt/struct.Layer.html#method.with_ansi), ansi is enabled by default when the crate `"ansi"` feature is enabled. If this is set to `true` and the flag is NOT enabled, then `with_ansi` panics

this is a roundabout way of saying that probably `log_color` should be ignored configs, and the recommendation to use `NO_COLOR` should be documented

ZanCorDX (approved):

PR #840, opened 2025-12-17
New metric for the max clickhouse backup size

ZanCorDX:
## 📝 Summary

This metric allows to automate alerts without hardcoded values.

* [X] Run `make lint`
* [X] Run `make test`
* [X] <-------<< Shot an arrow into the check list.


dvush (approved):
I am surprised that we need to add config values to the dashboard but if you want why not.

PR #842, opened 2025-12-19
refactor: reduce allocations in the sim task

prestwich:
## 📝 Summary

Reduce cloning in the simulator task. This minimizes and defers new allocations. It is both code cleanup and perf. There are no logic changes.

* [x] Run `make lint`
* [x] Run `make test`

### DIFF
--- a/crates/rbuilder/src/building/sim.rs
+++ b/crates/rbuilder/src/building/sim.rs
@@ -20,12 +20,13 @@ use rbuilder_primitives::{Order, OrderId, SimulatedOrder};
 use reth_errors::ProviderError;
 use reth_provider::StateProvider;
 use std::{
+    borrow::Cow,
     cmp::{max, min, Ordering},
     collections::hash_map::Entry,
     sync::Arc,
     time::{Duration, Instant},
 };
-use tracing::{error, trace};
+use tracing::{error, instrument, trace};
 
 #[derive(Debug)]
 #[allow(clippy::large_enum_variant)]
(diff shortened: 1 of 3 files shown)


crates/rbuilder/src/building/sim.rs
```diff
@@ -472,7 +475,7 @@ pub fn simulate_order_using_fork<Tracer: SimulationTracer>(
             let new_nonces = res.nonces_updated.into_iter().collect::<Vec<_>>();
             Ok(OrderSimResult::Success(
                 Arc::new(SimulatedOrder {
-                    order,
+                    order: order.into_owned(),
```
prestwich, line 478: using the Cow allows us to defer the clone to here. This means the clone is eliminated in the (presumably common) case that simulation fails

crates/rbuilder/src/building/sim.rs
```diff
@@ -434,8 +437,8 @@ pub fn simulate_order(
 
 /// Simulates order (including parent (those needed to reach proper nonces) orders) using a precreated fork
 pub fn simulate_order_using_fork<Tracer: SimulationTracer>(
-    parent_orders: Vec<Order>,
-    order: Order,
+    parent_orders: &[Order],
```
prestwich, line 440: using the reference allows us to not clone the parents, saving a potentially _very_ large allocation

metachris:
Thanks for the contribution, James! Most of the team is currently on vacation, and will be around again next week.

ZanCorDX:
Thanks for the work but we are about to change Order for Arc<Order> (since Order is something that almost never should be modified) so there is no need for introducing Cow.

prestwich:
Whether the Order is Arced or not, cloning the Vec requires an allocation which should be removed

PR #843, opened 2025-12-19
Refactor: fix docs, clean up code quality

prestwich:
## 📝 Summary

No longer based on #842, which has been closed

Improves and corrects documentation in the live_builder, particularly around Order propagation through the OrderPool. Functionality does not change. Most significant API change is around `BlobTypeOrderFilter`, by moving constructors from free functions to assoc functions

- Improve and correct `OrderPool` documentation
- Add `Copy` impl for `ReplacementData<KeyType>`
- Add missing `Copy` impl for `NonceKey`
- make implementation of `BlobTypeOrderFilter` construction idiomatic for Rust
- add missing `trace!` event to `BlobTypeOrderFilter` when filtering an order
- add `rule_name` to `BlobTypeOrderFilter` to ensure behavior is clear in tracing event
- add `From` impls for `ReplaceableOrderEvent`, and remove some relevant `.clone` invocations
 - update `AutoRemovingOrderPoolSubscriptionId` to use `Weak` to prevent them from holding `OrderPool` memory alloc for dying subscriptions
 - use `FuturesUnordered` instead of `Vec<JoinHandle<()>>` in the order pool to ensure parallelization of handle reaping 

* [x] Run `make lint`
* [x] Run `make test`


crates/rbuilder/src/live_builder/order_input/order_replacement_manager.rs
```diff
@@ -4,13 +4,20 @@ use rbuilder_primitives::{BundleReplacementData, BundleReplacementKey, Order, Or
...
+/// low level orderflow data via [`ReplaceableOrderSink`] and forwarding to an
+/// [`OrderSink`].
+///
+/// The `OrderReplacementManager` works for a single block.
+///
+/// IMPORTANT: Due to infra problems, notifications may arrive out of order.
+/// Cancels are treated as highest-priority, and after that we must always
+/// honor the replacement with highest `sequence_number`.
+///
+/// Although all the structs and fields say "bundle" we always refer to Bundle
```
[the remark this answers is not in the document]
prestwich: copilot this error is fixeeed in the diff

crates/rbuilder/src/live_builder/order_input/blob_type_order_filter.rs
```diff
@@ -1,62 +1,82 @@
...
+        Self::new(sink, "fusaka", fusaka)
+    }
+
+    /// Filters out [EIP-7594] style blobs, allowing only [EIP-4844] style.
+    ///
+    /// [EIP-4844]: https://eips.ethereum.org/EIPS/eip-4844
+    /// [EIP-7594]: https://eips.ethereum.org/EIPS/eip-7594
+    pub const fn new_pre_fusaka(sink: Box<dyn ReplaceableOrderSink>) -> Self {
+        fn pre_fusaka(tx: &TransactionSignedEcRecoveredWithBlobs) -> bool {
+            !tx.as_ref().is_eip4844() || tx.blobs_sidecar.is_eip4844()
```
prestwich, line 72: this logic relies on 4844 validity condition that transactions cannot have 0 blobs

See:
https://eips.ethereum.org/EIPS/eip-4844#blob-transaction:~:text=Execution%20layer%20validation,-On

crates/rbuilder-primitives/src/lib.rs
```diff
@@ -439,13 +441,17 @@ impl FakeSidecar for BlobTransactionSidecar {
 }
 
 /// First idea to handle blobs, might change.
-/// Don't like the fact that blobs_sidecar exists no matter if Recovered<TransactionSigned> contains a non blob tx.
-/// Great effort was put in avoiding simple access to the internal tx so we don't accidentally leak information on logs (particularly the tx sign).
+///
+/// Don't like the fact that blobs_sidecar exists no matter if
+/// [`Recovered<TransactionSigned>`] contains a non blob tx.
+///
+/// Great effort was put in avoiding simple access to the internal tx so we
```
prestwich, line 448: is this comment accurate? there is an `AsRef<Recovered<TransactionSigned>>` implementation that allows direct access to the internal tx

ZanCorDX: You are right about that. We could try to remove that (can_execute_with_block_base_fee needs it).
Maybe we can change the comment to say "be carefull with leaking tx internal info, this is the reason there is no getter for it"

ZanCorDX (approved):

PR #864, opened 2026-01-20
Stop submitting bids on clickhouse problems.

ZanCorDX:
## 📝 Summary

This PR introduces a trait RelaySubmissionPolicy to stop submitting on any problem.
I decided to cut the flow at the very end of the chain (I know it's inefficient) to play it super safe just in case any module is expecting new slot/block info periodically (eg: bidding services, caching stuff).
A new field was added to BuiltBlocksClickhouseConfig:

/// If set must be < disk_max_size_mb.
/// If the disk backup size is greater than this value, clickhouse will ask we stop submitting blocks.
pub disk_max_size_to_submit_bids_to_relays_mb: Option<u64>,

New metric clickhouse_disk_backup_max_size_to_submit_bids_to_relays_bytes with this value.

* [X] Run `make lint`
* [X] Run `make test`

### DIFF
--- a/crates/rbuilder/src/live_builder/block_output/relay_submit.rs
+++ b/crates/rbuilder/src/live_builder/block_output/relay_submit.rs
@@ -647,7 +647,7 @@ impl RelaySubmitSinkFactory {
     ) -> Box<dyn MultiRelayBlockBuildingSink> {
         // If submission is disabled, return a sink that throws away the blocks.
         if !self.submission_policy.should_submit() {
-            warn!("Submission is disabled by submission_policy, throwing away the blocks");
+            error!("Submission is disabled by submission_policy, throwing away the blocks");
             return Box::new(NullMultiRelayBlockBuildingSink {});
         }
         // Collect all relays to submit to.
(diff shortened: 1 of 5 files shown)


crates/rbuilder/src/live_builder/block_output/relay_submit.rs
```diff
@@ -616,6 +645,11 @@ impl RelaySubmitSinkFactory {
         slot_data: MevBoostSlotData,
         cancel: CancellationToken,
     ) -> Box<dyn MultiRelayBlockBuildingSink> {
+        // If submission is disabled, return a sink that throws away the blocks.
+        if !self.submission_policy.should_submit() {
+            warn!("Submission is disabled by submission_policy, throwing away the blocks");
```
dvush: I think this should be error and we need to have a metric for this, if its increasing we alert

dvush (approved):
Approved for faster merging, but check out my comment

crates/rbuilder/src/live_builder/block_output/relay_submit.rs
ZanCorDX: I didn't put this as an error since the real errors are "failed to commit bundle to clickhouse from backup" and a LOT more (we are not going to miss it). I saw this as a normal behavior triggered by the other errors, but I'll make it error to play it safer.
I though about the metric but we already have initiated_submissions going to 0.

PR #869, opened 2026-01-26
CI: Docker build for ARM64 too

metachris:
Add ARM64 as target platform for Docker builds.


bakhtin (approved):

ZanCorDX:
Took more than 2hs to build.

PR #903, opened 2026-03-22
perf: bypass precompile cache (0% hit rate)

vbuilder69420:
## Summary

- Bypass the `PrecompileCache` entirely — profiling shows **0 hits across 177,614 calls**
- The cache key includes `gas_limit` which varies per call, preventing any hits
- Each miss incurs a mutex lock + `Bytes` clone + HashMap lookup + second mutex lock + LruCache insert

## Motivation

Prometheus metrics from builder-lab profiling:
```
simulation_precompile_cache_hits 0
simulation_precompile_cache_misses 177614
```

The cache key is `(SpecId, Bytes, u64)` where the `u64` is `gas_limit`. Since `gas_limit` varies per precompile call (it's the remaining gas, not a fixed value), identical precompile inputs with different remaining gas produce different cache keys, preventing hits entirely.

## Benchmark Results

Tested with builder-lab, 100 TPS contender, 60s profiling window, stacked on top of the AccessListInspector removal (#902):

| Metric | Before | After | Change |
|--------|--------|-------|--------|
| Block fill time (p50) | 57.2ms | 53.3ms | **-6.8%** |
| Block fill time (p95) | 94.9ms | 80.7ms | **-15.0%** |
| E2E latency (p50) | 59ms | 55ms | **-6.8%** |
| E2E latency (p95) | 101ms | 84ms | **-16.8%** |

## Future improvement

Rather than bypassing the cache, it could be made effective by removing `gas_limit` from the cache key — precompile results don't depend on the gas limit passed to them (they either succeed within their fixed gas cost or fail). This would require verifying that all precompile implementations are indeed gas-limit-independent.

## Test plan

- [x] Blocks built and submitted correctly at 100 TPS
- [x] Confirmed 0% cache hit rate via Prometheus metrics
- [x] Profiling shows reduced overhead
- [ ] Integration tests pass

🤖 Generated with [Claude Code](https://claude.com/claude-code)

### DIFF
--- a/crates/rbuilder-primitives/src/evm_inspector.rs
+++ b/crates/rbuilder-primitives/src/evm_inspector.rs
@@ -1,7 +1,6 @@
 use ahash::HashMap;
 use alloy_consensus::Transaction;
 use alloy_primitives::{Address, B256, U256};
-use alloy_rpc_types::AccessList;
 use reth_primitives::{Recovered, TransactionSigned};
 use revm::{
     bytecode::opcode,
@@ -10,7 +9,6 @@ use revm::{
     interpreter::{interpreter_types::Jumps, CallInputs, CallOutcome, Interpreter},
     Inspector,
 };
-use revm_inspectors::access_list::AccessListInspector;
 
 #[derive(Debug, Clone, PartialEq, Eq, Hash, PartialOrd, Ord)]
 pub struct SlotKey {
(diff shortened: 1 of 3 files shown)


dmarzzz:
we should not merge this lol

vbuilder69420:
Reopening — tested with complex DeFi workload (AMM swaps, lending, oracle updates, liquidations via shared economy contender scenario). Still 0 hits across 25,573 precompile calls. The gas_limit in the cache key prevents any hits regardless of workload complexity. The cache is pure overhead.

PR #907, opened 2026-03-23
fix: ClickHouse backup drain loop — batch merging and bug fixes

fsodre:
## Summary

- **Continuous drain**: The backup drain loop now processes all backed-up batches in a single tick cycle instead of one batch per tick, fixing slow recovery after ClickHouse outages observed in production
- **Batch merging**: Multiple small batches are merged into a single ClickHouse commit (up to 65,536 rows), reducing HTTP round-trips from N to ~N/batch_size
- **Disk key preservation**: Fixed a bug where disk source keys were lost when a merged commit failed and was re-stored for retry, causing duplicate rows in ClickHouse after recovery
- **Memory pressure safety**: Fixed a bug where merged retry batches (containing disk-sourced data) could be silently dropped under memory pressure, causing permanent data loss
- **Backoff correctness**: The backoff timer is now only reset when a drain completes fully, not on each individual successful chunk within a drain cycle
- **Compilation fix**: Added missing `derive` feature to `strum` dependency

## Test plan

- [x] 9 unit tests covering drain behavior, batch merging, row limits, and bug regressions
- [x] 2 integration tests against real ClickHouse (Docker) verifying row counts and zero duplicates
- [x] Outage simulation test: stop ClickHouse mid-drain, accumulate batches, restart, verify full recovery with no duplicates
- [ ] Run `./scripts/test-clickhouse-backup-drain.sh` (starts Docker ClickHouse, runs all tests, verifies data, cleans up)
- [ ] Run `./scripts/test-clickhouse-backup-drain.sh --basic` to skip the outage simulation if Docker stop/start is slow

🤖 Generated with [Claude Code](https://claude.com/claude-code)

### DIFF
--- a/crates/rbuilder-utils/Cargo.toml
+++ b/crates/rbuilder-utils/Cargo.toml
@@ -22,7 +22,7 @@ serde_json.workspace = true
 # alloy
 alloy-primitives.workspace = true
 
-strum = "0.27"
+strum = { version = "0.27", features = ["derive"] }
 tokio = { version = "1.40.0", default-features = false, features = [
   "sync",
   "time",
--- a/crates/rbuilder-utils/src/clickhouse/backup/mod.rs
+++ b/crates/rbuilder-utils/src/clickhouse/backup/mod.rs
@@ -764,9 +764,11 @@ impl<T: ClickhouseRowExt, MetricsType: Metrics> Backup<T, MetricsType> {
             if let Err(e) = self.inserter.force_commit().await {
                 tracing::error!(target: TARGET, order = T::TABLE_NAME, ?e, "failed to commit cached backup to CH during shutdown, trying disk");
                 MetricsType::increment_commit_failures(e.to_string());
-            }
-            if let Err(e) = self.disk_backup.save(&cached.commit) {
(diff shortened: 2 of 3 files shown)


fsodre:
### Code review

Found 1 issue:

1. **Disk entries re-read repeatedly in the gather loop, causing duplicate rows in ClickHouse.** `DiskBackup::retrieve_oldest()` is a read-only operation (takes `&self`, never calls `delete`). Once memory is exhausted and `last_cached` is empty, the inner `while` gather loop calls `self.retrieve_oldest()` repeatedly, and each call returns the **same** disk entry because it was never consumed. The loop extends `merged_rows` with the same rows over and over until `max_rows_per_commit` is reached, then commits all those duplicates to ClickHouse. The original code was safe because it processed one batch per tick and purged immediately after success. The new gather-then-commit pattern breaks that invariant.

https://github.com/flashbots/rbuilder/blob/40ce0461882c07b7a9694e8db35a14c621f8fd95/crates/rbuilder-utils/src/clickhouse/backup/mod.rs#L721-L732

🤖 Generated with [Claude Code](https://claude.ai/code)

<sub>- If this code review was useful, please react with 👍. Otherwise, react with 👎.</sub>

PR #910, opened 2026-03-31
feat: more paranoid reth db checks

dvush:
Now we check reth db account trie table for correct state root of the parent instead of checking latest block.

## 💡 Motivation and Context

---

* [ ] Run `make lint`
* [ ] Run `make test`


ZanCorDX (approved):

ZanCorDX:
Failed when deployed

PR #855, opened 2026-01-13
add builder server for epbs 

faheelsattar:
## 📝 Summary

- lays ground work for epbs builder server.
- `get_bids` endpoint for retrieving bids for the proposer

## 💡 Motivation and Context

---

## ✅ I have completed the following steps:

* [ ] Run `make lint`
* [ ] Run `make test`
* [ ] Added tests (if applicable)

PR #928, opened 2026-06-30
Upgrade reth 1.11.3 -> 2.2.0

faheelsattar:
## 📝 Summary

Bumps reth version from 1.11.3 to 2.2.0, and alloy dependencies.

## 💡 Motivation and Context

As part of adding support for epbs in rbuilder we were required to upgrade reth to 2.0. We initially added it via https://github.com/flashbots/rbuilder/pull/855, but later decided to separate out the version bump in its own PR.

* [x] Run `make lint`
* [x] Run `make test`


crates/rbuilder/src/building/mod.rs
```diff
@@ -172,20 +206,21 @@ impl BlockBuildingContext {
                     suggested_fee_recipient: attributes.suggested_fee_recipient(),
                     prev_randao: attributes.prev_randao(),
                     gas_limit,
-                    withdrawals: Some(attributes.withdrawals.clone()),
-                    parent_beacon_block_root: attributes.parent_beacon_block_root,
-                    extra_data: extra_data.clone().into(),
+                    withdrawals: Some(attributes.withdrawals()),
+                    parent_beacon_block_root: attributes.parent_beacon_block_root(),
+                    extra_data: Bytes::new(),
```
ZanCorDX: Why did you remove the extra data? Probably it's not really needed now but maybe it's safer for future changes to keep the real one?

crates/rbuilder/src/building/payout_tx.rs
```diff
@@ -97,11 +98,7 @@ where
 
     let res = evm.transact(&tx)?;
     match res.result {
-        ExecutionResult::Success {
-            gas_used,
-            gas_refunded,
-            ..
-        } => Ok(Some(gas_used + gas_refunded)),
+        ExecutionResult::Success { .. } => Ok(Some(res.result.tx_gas_used())),
```
ZanCorDX: You removed the gas_refunded why?

faheelsattar: the new revm changed `ExecutionResult::Success` to carry a single `ResultGas` instead of seperate gas_used and gas_refunded fields. Therefore i think instead of using just tx_gas_used, I can use total_gas_spent, which returns the total gas spent inside execution before any refunds

ZanCorDX: I think you should us total_gas_spent or add refunded gas.
Notice:
```
    #[inline]
    pub const fn tx_gas_used(&self) -> u64 {
        // consiste of regular and state gas.
        let total_gas_spent = self.total_gas_spent();
        // from total gas subtract the refunded gas. Refunded is capped by 20% of total gas spent.
        let tx_gas_refunded = total_gas_spent.saturating_sub(self.inner_refunded());
        max(tx_gas_refunded, self.floor_gas())
    }
```
As far as I understand, a tx is executed consuming up to the gaslimit set and *after* execution gas refunds are performed. So if the extra "refund gas" is not added to the limits the tx will fail. Do I have this wrong?

ZanCorDX (approved):

PR #930, opened 2026-07-08
Upgrade reth 1.11.3 -> 2.2.0 (copy to fix CI)

ZanCorDX:
## 📝 Summary
Original work by @faheelsattar 
Bumps reth version from 1.11.3 to 2.2.0, and alloy dependencies.

## 💡 Motivation and Context

As part of adding support for epbs in rbuilder we were required to upgrade reth to 2.0. We initially added it via https://github.com/flashbots/rbuilder/pull/855, but later decided to separate out the version bump in its own PR.

* [x] Run `make lint`
* [x] Run `make test`


aristide-bayard:
Hi guys how is this pr going? Would be great to be able to upgrade to reth v2 to work with the builder!

